PR #7, opened 2025-05-23
Views

AbdelrhmanBassiouny:
This pull request introduces a rule-based classification system for the `semantic_world` module, focusing on classifying "views" within a `World` object. The most important changes include the addition of new classifiers, integration of a ripple-down rules (RDR) framework.

RDR development/fitting is done through introducing  more test cases or increasing the scope of a test case. So basically you will go to test_views.py, you will load the correct rules and add new rules by adding a new type of view to the inputs of the fit_views_and_get_rdr() function, or you can improve on rules of current views but setting update_current_views=True. An example is as follows, in the test_views.py:

```python
world = self.kitchen_parser.parse()
world.validate()

# To add a new view type (not limited to two, this is just an example :))
rdr = self.fit_views_and_get_rdr(world, [MyNewViewType1, MyNewViewType2])

# Or if you want to update an existing view type (not limited to two, this is just an example :))
rdr = self.fit_views_and_get_rdr(world, [OldViewType1, OldViewType2], update_existing_views=True)

found_views = rdr.classify(world)['views]

self.assertTrue(any([type(v) is MyNewViewType for v in found_views]))
```
And then be ready to be prompted to write a rule :smiley: .

Currently the generated files from the rdr is in src/semantic_world/views/views_classifiers. So ig you just want to use them as is, then just import the classifier and use it like this:

```python
from semantic_world.views.view_classifiers import world_rdr

my_world = URDFParser(urdf_path).parse()
found_views = world_rdr.classify(world)["views"]

print(found_views)
```


src/semantic_world/views/view_classifiers/world_views_mcrdr_of_grdr_defs.py
tomsch420, line 1: formatting

test/test_views/test_views.py
```diff
@@ -0,0 +1,99 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls.kitchen_parser = URDFParser(cls.kitchen)
+        cls.apartment_parser = URDFParser(cls.apartment)
+        cls.app = QApplication(sys.argv)
+        cls.viewer = RDRCaseViewer(save_file=cls.views_json_dir)
+
+    def test_id(self):
+        v1 = Handle(1)
+        print(hash(v1))
```
tomsch420, line 40: remove

tomsch420 (changes requested):

src/semantic_world/views/view_classifiers/world_views_mcrdr_of_grdr_defs.py
```diff
@@ -0,0 +1,93 @@
...
+from semantic_world.connections import PrismaticConnection
+from semantic_world.views.views import Drawer
+from semantic_world.views.views import Handle
+from typing import Union
+from semantic_world.views.views import Cabinet
+from semantic_world.world import World
+from semantic_world.views.views import Container
+from typing_extensions import Union
+
+
```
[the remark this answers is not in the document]
AbdelrhmanBassiouny: this is auto generated.

tomsch420 (approved):

tomsch420 (approved):

PR #9, opened 2025-05-30
New File for Corner Cases of Rules on Views.

AbdelrhmanBassiouny:
A new auto generated file that stores case factory method and producing scenario (i.e. the test case that resulted in this rule).

This helps in maintining the rule base and will help in the future to compare/merge different rule bases that produce same types of conclusions against the same corner cases.

Refactored test_views to have a separate test case for each view for more accurate tracking of scenarios that produced views.


src/semantic_world/views/world_rdr/world_views_mcrdr_cases.py
tomsch420, line 1: not a problem but this file is vwery messy due to the many repeated imports.

src/semantic_world/views/world_rdr/world_views_mcrdr_defs.py
```diff
@@ -1,14 +1,14 @@
+from typing import List
+from semantic_world.connections import PrismaticConnection
+from semantic_world.views.views import Cabinet
+from semantic_world.connections import RevoluteConnection
```
tomsch420, line 4: absolute imports?

AbdelrhmanBassiouny: was not easy to do , but done haha.

src/semantic_world/views/world_rdr/world_rdr.py
```diff
@@ -1,10 +1,9 @@
...
+from ...world import World
 from ripple_down_rules.datastructures.case import Case, create_case
-from semantic_world.world import World
+from ripple_down_rules.helpers import general_rdr_classify
+from typing import Any, Dict
 from . import world_views_mcrdr as views_classifier
 
-
 classifiers_dict = dict()
 classifiers_dict['views'] = views_classifier
```
tomsch420, line 8: what does this one do? Btw, classes can also be keys

tomsch420 (changes requested):

src/semantic_world/views/world_rdr/world_rdr.py
AbdelrhmanBassiouny: this is the main rdr classifier, since this is a GenralRDR it can classify many different types/attributes for a case, here it only has a classifier for world views, but in general it could have other world attribute classifiers as well.

Yes I know I could but a class instead of the attribute name but sometimes the different attributes have the same type and this would be a problem, for examples to attributes to type bool.

AbdelrhmanBassiouny: and this is also a generated script, you can import and use this for classification, and does not use rdrs here, it is just a normal python function, by has the same behaviour as if you were classifying using the rdr instance.

AbdelrhmanBassiouny: see test_generated_views in the test_views.py

PR #13, opened 2025-06-28
Viz publisher

Tigul:
Adds the VizMarkerPublisher, the publisher takes a world and node and publishes the contents of the world as VizMarkers which can be seen in RViz2. 

The Publisher is only available in ROS2

This PR also improves the URDFParser to parse the collision information of URDFs.

### DIFF
--- a/src/semantic_world/adapters/viz_marker.py
+++ b/src/semantic_world/adapters/viz_marker.py
@@ -64,7 +64,7 @@ class VizMarkerPublisher:
                 msg.ns = body.name.name
                 msg.id = i
                 msg.action = Marker.ADD
-                msg.pose = self.transform_to_pose(self.world.compute_forward_kinematics_np(self.world.root, body))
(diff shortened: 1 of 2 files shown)


tomsch420 (approved):
Looks good, is the Publisher converting to the right ros Type? Like int to float and so on

src/semantic_world/adapters/viz_marker.py
```diff
@@ -0,0 +1,108 @@
...
+
+        :return: An Array of Visualization Marker
+        """
+        marker_array = MarkerArray()
+        for body in self.world.bodies:
+            for i, collision in enumerate(body.collision):
+                msg = Marker()
+                msg.header.frame_id = self.reference_frame
+                msg.ns = body.name.name
+                msg.id = i
```
tomsch420, line 65: This Looks Like it works only for one body

tomsch420 (approved):

PR #27, opened 2025-07-11
Gave every WorldEntity a name and added add/get view 

ichumuh:
I noticed that basically every world entity subclass has a name, while I was creating new views. So why not simply give every entity a name?

without a name it is also quite difficult to retrieve a view from the world

### DIFF
--- a/src/semantic_world/world.py
+++ b/src/semantic_world/world.py
@@ -406,6 +406,51 @@ class World:
         connection._world = self
         self.kinematic_structure.add_edge(connection.parent.index, connection.child.index, connection)
 
+    def add_view(self, view: View) -> None:
+        """
+        Adds a view to the current list of views if it doesn't already exist. Ensures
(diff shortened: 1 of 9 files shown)


src/semantic_world/world.py
```diff
@@ -406,6 +406,51 @@ def add_connection(self, connection: Connection) -> None:
...
+        try:
+            self.get_view_by_name(view.name)
+        except ValueError:
+            pass
+        else:
+            raise ValueError(f"View with name {view.name} already exists.")
+        view._world = self
+        self.views.append(view)
+
+    def get_view_by_name(self, name: Union[str, PrefixedName]) -> View:
```
tomsch420, line 429: id prefer return Optional[View] instead of the raising here.

tomsch420 (approved):
For this to work you also have to reexecute generate_orm.py

PR #28, opened 2025-07-14
added style guide for transformation matrices

ichumuh:
related to https://github.com/cram2/semantic_world/pull/26

### DIFF
--- a/doc/_toc.yml
+++ b/doc/_toc.yml
@@ -7,3 +7,4 @@ chapters:
 - file: bibliography
 - file: kinematic_world
 - file: autoapi/index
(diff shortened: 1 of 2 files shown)


LucaKro (commented):
This is really nice, I think this project benefits greatly from these kinds of style guides! The only thing I would comment on is that I think adding a few small practical examples could also help, maybe a running example that is used throughout the guide

ichumuh:
i did make small examples. My intention wasn't to make a guide to how homogeneous transformation matrices work. if you want that, then it would probably be better to search for a guide online and link it here

PR #34, opened 2025-07-15
Added parsing of filepaths in URDFs

Tigul:
# Description 
Adds a parsing function which looks up the real path of ROS package paths in URDFs. 

The ROS function is only imported if the URDF contains ROS package paths.


tomsch420 (approved):

tomsch420:
Does this need a ROS docker to work?

Tigul:
Jep either that or we use another URDF for testing which does not have meshes like the pr2_kinematic_tree.urdf

PR #39, opened 2025-07-17
Merge pose

Tigul:
# Description
This PR fixes the merge_worlds method by creating a connection between the roots of both world thus resulting in a valid world model. (Feedback on the signatur of merge_worlds is appriciated, we could also  do merge_worlds(self, other, self_mount_body, connection_class))

It also adds a method merge_worlds_with_pose which merges both worlds and creates a 6DOF connection with the pose as origin.

### DIFF
--- a/src/semantic_world/world.py
+++ b/src/semantic_world/world.py
@@ -6,7 +6,7 @@ from copy import deepcopy
 from dataclasses import dataclass, field
 from enum import IntEnum
 from functools import wraps, lru_cache
-from typing import Dict, Tuple, OrderedDict, Union, Optional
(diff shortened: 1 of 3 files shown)


src/semantic_world/world.py
```diff
@@ -485,6 +491,21 @@ def merge_world(self, other: World) -> None:
             self.add_connection(connection)
         other.world_is_being_modified = False
 
+        connection = partial_connection or Connection6DoF(parent=self_root, child=other_root, _world=self)
+        self.add_connection(connection)
+
+    def merge_world_with_pose(self, other: World, pose: NpMatrix4x4) -> None:
```
ichumuh, line 497: merge_world_at_pose?

PR #40, opened 2025-07-17
use Body instead of prefixedname when possible

ichumuh:
- reference_frame and child_frame of spatial types are of type body
- refactored degrees of freedom
- includes https://github.com/cram2/semantic_world/pull/36


src/semantic_world/spatial_types/derivatives.py
```diff
@@ -19,4 +24,41 @@ def range(cls, start: Derivatives, stop: Derivatives, step: int = 1):
         """
         Includes stop!
         """
-        return [item for item in cls if start <= item <= stop][::step]
\ No newline at end of file
+        return [item for item in cls if start <= item <= stop][::step]
+
+
+@dataclass
+class DerivativeMap(Generic[T]):
```
tomsch420, line 31: doc

tomsch420 (approved):

src/semantic_world/spatial_types/derivatives.py
ichumuh: had to test if you read the code

PR #45, opened 2025-07-21
Robot view checkup

LucaKro:
Removed RobotBody, Rewrote PR2 view class, fixed faulty aggregation of bodies when dealing with custom properties and private fields.
added field "primary_prefix" to world because otherwise its really hacky to check which prefix should be assigned to the views, or the views just don't have a prefix which is also weird. alternative could be to add "prefix" param to from_world method.


LucaKro:
Cant ping @ichumuh as a reviewer

src/semantic_world/world_entity.py
```diff
@@ -248,18 +256,45 @@ class View(WorldEntity):
     This class can hold references to certain bodies that gain meaning in this context.
     """
 
-    @property
-    def aggregated_bodies(self) -> Set[Body]:
+    @staticmethod
+    def _is_relevant(x: object) -> bool:
```
tomsch420, line 260: the method name is not really reflecting whats happening here.

ichumuh: i agree, x is also not a good argument name

src/semantic_world/world_entity.py
```diff
@@ -268,26 +303,27 @@ def aggregated_bodies(self) -> Set[Body]:
...
+                    stack.extend(v for v in obj.values() if self._is_relevant(v))
+
                 case Iterable() if not isinstance(obj, (str, bytes, bytearray)):
-                    stack.extend(obj)
+                    stack.extend(v for v in obj if self._is_relevant(v))
+
         return bodies
 
+    @property
+    def aggregated_bodies(self) -> Set[Body]:
```
ichumuh, line 321: why not `.bodies`?

ichumuh: would make it the same interface the world uses

src/semantic_world/world.py
```diff
@@ -260,6 +260,11 @@ class World:
     Is set to True, when a function with @modifies_world is called or world.modify_world context is used.
     """
 
+    primary_prefix: Optional[str] = None
+    """
+    Prefix used for the world. Acts as default namespace for all bodies and views in the world which do not have a prefix.
+    """
```
ichumuh, line 266: hm, idk about the primary prefix. how about we give the world a name too, like everything else, and use that as the primary prefix?

ichumuh (changes requested):

src/semantic_world/robots.py
```diff
@@ -96,9 +112,18 @@ class Neck(KinematicChain):
     Represents a special kinematic chain to identify the different bodys of the neck, which is useful to calculate
     for example "LookAt" joint states without needing IK
     """
-    roll_body: Optional[RobotBody] = None
-    pitch_body: Optional[RobotBody] = None
-    yaw_body: Optional[RobotBody] = None
+    roll_body: Optional[Body] = None
```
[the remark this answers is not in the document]
LucaKro: sometimes IK is a bit overkill for computing the joint positions, for example in pycram we calculate the move head joint positions by calculating for example the angle of elevation. but to be able to do that we need to know which joint is, for example, the joint doing the pitch rotation in a 2 joint robot neck. This is what this is for. 
>  i think it would make more sense to have the connections, like roll_connection, and ask them for the bodies if you actually need them
i guess we could also do that, but i think this is a bit weird tbh. If we use the body, then during those calculations we can just query the parent connection. We have a intuitive way of getting the bodies too: we can just query the correct body by using its name. 
if we use the connections instead, the user still has to know the corresponding body name, but instead of using to query the body they query the parent connection of the body.
one can also query the connection directly, but we are not actually using the joint names of the urdf in our graph, so instead of intuitively looking at the urdf and taking the correct joint name (which i agree would probably be the best solution), the user now needs to be familiar with how the semantic world names connections internally, and then for example query 'head_pan_link_T_head_tilt_link', which requires even more information on the user side (this also means steeper learning curve).
I think from our structure, it makes most sense to do it the way I am doing it here, since we do not get around the user having to associate the "body" with the rotation. One alternative would be to add a property "urdf_name" or something similar (since we already have prefixed names like the one above attached to each connection), which would be the name actual name of the joint in the urdf, for example "head_pan_joint".

src/semantic_world/world.py
LucaKro: I mean that would just be a change in name right? That makes sense to me, tho I will first look at the naming changes bass merged, this may be obsolete. If its not ill give the world a name instead

src/semantic_world/robots.py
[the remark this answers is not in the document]
ichumuh: 1. my every has name PR should have given connections generated from urdf the joint name from the urdf
2. is the only use case a simplified IK? which is then sent as a joint goal to giskard? if so then you shouldn't need it. giskard can describe pointing goals, so no need for you to do giskards job. and if you really need the IK for some other reason, this simplification should be hidden behind the compute IK feature of the world.

src/semantic_world/robots.py
```diff
@@ -1,85 +1,324 @@
...
+        """
+        raise NotImplementedError("This method should be implemented in subclasses.")
+
+    @abstractmethod
+    def add_to_robot(self, robot: AbstractRobot):
+        """
+        Adds this view to the robot's collection of kinematic chains.
+        This method should be implemented in subclasses to ensure the view is properly added to the robot.
+
+        :param robot: The robot to which this view should be added.
```
ichumuh, line 43: i don't get it. why kinematic chains?

src/semantic_world/robots.py
```diff
@@ -1,85 +1,324 @@
...
 @dataclass
-class RobotView(RootedView):
+class KinematicChain(RobotView, ABC):
     """
-    Represents a collection of connected robot bodies, starting from a root body, and ending in a unspecified collection
-    of tip bodies.
+    Abstract base class for kinematic chain in a robot, starting from a root body, and ending in a specific tip body.
+    A kinematic chain can contain both a manipulator and sensors at the same time.
+    """
+    tip_body: Body = field(default_factory=Body)
```
ichumuh, line 54: naming this `tip` would be more consistent with the rest of the codebase no?

src/semantic_world/robots.py
```diff
@@ -1,85 +1,324 @@
...
+    """
+    The tip body of the kinematic chain, which is the last body in the chain.
+    """
+
+    manipulator: Optional[Manipulator] = None
+    """
+    The manipulator of the kinematic chain, if it exists. This is usually a gripper or similar device.
```
ichumuh, line 61: are manipulator and sensor expected to be at/after tip?

src/semantic_world/robots.py
```diff
@@ -1,85 +1,324 @@
...
+        Adds this manipulator to the robot's collection of manipulators.
+        This method is called when the manipulator is created and should be implemented in subclasses.
+
+        :param robot: The robot to which this manipulator should be added.
+        """
+        if self._robot is not None and self._robot != robot:
+            raise ValueError(f"Manipulator {self.name} is already part of another robot: {self._robot.name}.")
+        if self._robot is not None and self._robot == robot:
+            return
+        self._robot = robot
```
ichumuh, line 173: you are not adding it as a manipulator to the robot, is that a bug?

src/semantic_world/robots.py
```diff
@@ -89,28 +328,143 @@ class Camera(Sensor):
...
 @dataclass
 class Torso(KinematicChain):
     """
     A Torso is a kinematic chain connecting the base of the robot with a collection of other kinematic chains.
     """
-    kinematic_chains: List[KinematicChain] = field(default_factory=list)
+    kinematic_chains: Set[KinematicChain] = field(default_factory=set)
     """
     A collection of kinematic chains, such as sensor chains or manipulation chains, that are connected to the torso.
     """
```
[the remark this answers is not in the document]
ichumuh: like if you care about arm, you'd do robot.arms, no?

ichumuh (changes requested):

src/semantic_world/robots.py
LucaKro: add_to_robot naming is confusing, im changing it to "assign_to_robot" and update the docstrings accordingly

src/semantic_world/robots.py
```diff
@@ -1,85 +1,324 @@
...
+        :return: A ParallelGripper object if the palm and tool frame bodies are found, otherwise raises ValueError.
+        """
+        if len(finger_bodys) != 2:
+            raise ValueError("Parallel gripper must have exactly two fingers")
+        finger, thumb = None, None
+        for index, (finger_root_name, finger_tip_name) in enumerate(finger_bodys):
+            if index == 0:
+                thumb = Finger.create(world, name + '_thumb', finger_root_name, finger_tip_name)
+            elif index == 1:
+                finger = Finger.create(world, name + '_finger', finger_root_name, finger_tip_name)
```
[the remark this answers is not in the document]
LucaKro: changed so that the fingers need to be created externally and passed to this method instead

src/semantic_world/robots.py
```diff
@@ -133,275 +487,162 @@ class AbstractRobot(RootedView):
...
+        """
+        Creates a robot view from the given world.
+        This method constructs the robot view by identifying and organizing the various semantic components of the robot,
+        such as manipulators, sensors, and kinematic chains. It is expected to be implemented in subclasses.
+
+        :param world: The world from which to create the robot view.
+
+        :return: A robot view.
+        """
+        raise NotImplementedError("This method should be implemented in subclasses.")
```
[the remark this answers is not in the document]
LucaKro: this could be useful in the future, but at the moment we only support one robot per type, since anything else will run into naming problems. Its a todo ill write down, but which I dont think should be addressed in this pr

src/semantic_world/robots.py
```diff
@@ -133,275 +487,162 @@ class AbstractRobot(RootedView):
...
+
+    def add_manipulator(self, manipulator: Manipulator):
+        """
+        Adds a manipulator to the robot's collection of manipulators.
+        """
+        self.manipulators.add(manipulator)
+        self._views.append(manipulator)
+        manipulator.add_to_robot(self)
```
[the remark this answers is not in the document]
LucaKro: I am changing _views to be sets too, this should then solve it right?

src/semantic_world/robots.py
```diff
@@ -133,275 +487,162 @@ class AbstractRobot(RootedView):
...
+        self.sensors.add(sensor)
+        self._views.append(sensor)
+        sensor.add_to_robot(self)
+
+    def add_torso(self, torso: Torso):
+        """
+        Adds a torso to the robot's collection of kinematic chains.
+        """
+        if self.torso is not None:
+            raise ValueError(f"Robot {self.name} already has a torso: {self.torso.name}.")
```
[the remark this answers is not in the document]
LucaKro: because right now thats how abstract robot defines a torso, and we dont have any case so far where a robot would have two torsos. I think we should add the support if we ever run into the problem in the future, until then its not necessary. 

also having two torsos would, in my opinion make the torso.kinematic_chain property more useful

src/semantic_world/world.py
```diff
@@ -471,11 +476,14 @@ def merge_world(self, other: World, root_connection: Connection = None) -> None:
...
         Merge a world into the existing one by merging degrees of freedom, states, connections, and bodies.
         This removes all bodies and connections from `other`.
+        The self does not have a primary prefix, it will be set to the primary prefix of the other world.
 
         :param other: The world to be added.
         :param root_connection: If provided, this connection will be used to connect the two worlds. Otherwise, a new Connection6DoF will be created
         :return: None
         """
+        if self.name is None:
+            self.name = other.name
```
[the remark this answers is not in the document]
LucaKro: you think the current name should just remain none then?

src/semantic_world/robots.py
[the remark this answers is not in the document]
ichumuh: oh that's why everything is a set, currently i don't see a reason why not. maybe we should also consider making bodies, connections, views of a world a set :thinking:

src/semantic_world/robots.py
LucaKro: atm there are no assumptions about where either are. what would you suggest?

src/semantic_world/world_entity.py
```diff
@@ -248,18 +258,45 @@ class View(WorldEntity):
...
+    @staticmethod
+    def _is_body_view_or_iterable(obj: object) -> bool:
+        """
+        Determines if an object is a Body, a View, or an Iterable (excluding strings and bytes).
         """
-        Recursively traverses the view and its attributes to find all bodies contained within it.
+        return (
+                isinstance(obj, (Body, View)) or
+                (isinstance(obj, Iterable) and not isinstance(obj, (str, bytes, bytearray)))
+        )
```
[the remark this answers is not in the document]
ichumuh: class attribute, sry. i mean moving it out of the class as free standing methods.

src/semantic_world/robots.py
ichumuh: looking at the pr2 arm, the forearm camera would be somewhere in the middle and the gripper at the end, which feels a bit random. but it depends what you want to use this information for i guess. maybe just write that the manipulators/sensors can be attached anywhere at the kinematic chain and leave it as it is?

src/semantic_world/world_entity.py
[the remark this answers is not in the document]
LucaKro: do I have them as free standing functions inside world_entity? I cannot put it into utils because id need to import body and views

src/semantic_world/robots.py
LucaKro: i think so too, as the views are only accessors, and if we need specific information one can still do kinematic_chain.manipulator.root or something like that

src/semantic_world/robots.py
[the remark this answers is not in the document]
LucaKro: thats fair, i didnt mean to imply that we agreed on anything formally or sth, but I did talk to people, and if there were complaints or people were not happy with anything i proposed, I did change it. 

I will remove it for now and also check for other cases where this may apply

ichumuh (approved):

tomsch420 (approved):

PR #48, opened 2025-07-30
Spatial types everywhere

ichumuh:
- includes https://github.com/cram2/semantic_world/pull/46
- worldentity generates a unique name based on hash, if no name is given
- replaced unitvector with a factory method for vector3
- replaced `world.compute_relative_pose` with `world.transform`, since the spatial types know their reference frame
- refactored point3, vector3 and quaternion to use x/y/z(/w) in `__init__` instead of data, because it is more readable and consistent with unitvector. `from_iterable` can be used for old behavior
- added a lot of tests for spatial types
- reverted `get_view_by_name_and_type` to `get_view_by_name`
- test_table_world needs to be fixed by @tomsch420  
- deleted test_views/pytest.ini, because this prevents pycharm from running individual tests in that folder. instead you should set -s in the run templates in pycharm as an additional pytest option


src/semantic_world/world_entity.py
```diff
@@ -359,25 +362,51 @@ def __post_init__(self):
...
+        Executes post-initialization logic based on the presence of a world attribute.
+        """
+        if self._world is None:
+            self._post_init_without_world()
+        else:
+            self._post_init_with_world()
+
+    def _post_init_with_world(self):
+        """
+        This method is invoked to initialize or perform additional setup operations
```
tomsch420, line 376: imperative

src/semantic_world/world_entity.py
```diff
@@ -359,25 +362,51 @@ def __post_init__(self):
...
+        that are required after the main initialization step. It is intended for
+        use cases involving world-related configurations or any specific setup
+        details required post object creation.
+        """
+        pass
+
+    def _post_init_without_world(self):
+        """
+        This method is intended for internal initialization processes and is not meant for external use.
```
tomsch420, line 385: imperative

src/semantic_world/world.py
```diff
@@ -840,20 +819,48 @@ def compute_forward_kinematics_np(self, root: Body, tip: Body) -> NpMatrix4x4:
         :param tip: Tip body to which the kinematics are computed.
         :return: Transformation matrix representing the relative pose of the tip body with respect to the root body.
         """
-        return self._fk_computer.compute_forward_kinematics_np(root, tip).copy()
+        return cas.TransformationMatrix(self.compute_forward_kinematics_np(root, tip))
+
+    def compute_forward_kinematics_np(self, root: Body, tip: Body) -> NpMatrix4x4:
+        """
+        Computes the forward kinematics from the root body to the tip body, root_T_tip and returns it as a 4x4 numpy ndarray.
```
tomsch420, line 826: imperative

src/semantic_world/world.py
```diff
@@ -840,20 +819,48 @@ def compute_forward_kinematics_np(self, root: Body, tip: Body) -> NpMatrix4x4:
...
-        :param target_body: The body to which the pose should be transformed
-        :param pose_body: The body which should be used as reference frame for the pose
-        :return: The pose relative to the target body.
+        return self._fk_computer.compute_forward_kinematics_np(root, tip).copy()
+
+    def transform(self, spatial_object: cas.SpatialType, target_frame: Body) -> cas.SpatialType:
         """
-        target_T_pose = self.compute_forward_kinematics_np(target_body, pose_body)
-        return target_T_pose @ pose
+        Transforms a given spatial object from its reference frame to a target frame.
```
tomsch420, line 839: imperative

tomsch420 (approved):

src/semantic_world/world_entity.py
ichumuh: i thought asking ai to write my docs strings is good enough? :(

src/semantic_world/connections.py
```diff
@@ -22,6 +24,7 @@ class Has1DOFState:
     """
 
     dof: DegreeOfFreedom
+    _world: World
```
LucaKro, line 27: why is this needed?

src/semantic_world/world_entity.py
```diff
@@ -61,6 +61,10 @@ class WorldEntity:
     The identifier for this world entity.
     """
 
+    def __post_init__(self):
+        if self.name is None:
+            self.name = PrefixedName(f"{self.__class__.__name__}_{hash(self)}")
```
LucaKro, line 66: we have an IDGenerator class for that, i think it makes sense to have this consistent throughout the codebase, and IDGenerator gives us nicer IDs compared to hashes

src/semantic_world/connections.py
ichumuh: its not strictly needed for the code to work. but it stops pycharm from complaining, because we use _world in the properties of this class

src/semantic_world/world_entity.py
ichumuh: i tried the id generator before, but it doesn't work here.
example:
you have a view called handle, which only has a body attribute. if you have two handles that refer to the same body, then the id generate will create two different names for both. the hash based generator will create the same name.

rdr actually autogenertes views and assumes that they are identical, if the attributes are identical. but with the idgenerator based version this wasn't the case, causing some rdr tests to never terminate.

PR #49, opened 2025-08-11
Mesh Adapter

Tigul:
# Description 
Added a simple adapter for parsing meshes to bodies, this uses trimesh so as long as a format is supported the parser can be extended to this format.


resources/stl/jeroen_cup.stl
tomsch420, line 1: uhhhhh do we want those in the git?

tomsch420 (approved):

PR #51, opened 2025-08-13
collision config

ichumuh:
- added collision configs to all bodies
- added collision detector base class

### DIFF
--- a/src/semantic_world/robots.py
+++ b/src/semantic_world/robots.py
@@ -439,14 +439,6 @@ class AbstractRobot(RootedView, ABC):
         self._views.add(kinematic_chain)
         kinematic_chain.assign_to_robot(self)
 
-    @cached_property
-    def unmovable_bodies_with_collision(self) -> Set[Body]:
-        result = set()
-        for body in self.bodies_with_enabled_collision:
-            if not self._world.is_controlled_connection_in_chain(self._world.root, body):
-                result.add(body)
-        return result
-
 @dataclass
 class PR2(AbstractRobot):
     """


src/semantic_world/adapters/urdf.py
```diff
@@ -21,7 +21,7 @@
     'fixed': FixedConnection}
 
 
-def urdf_joint_to_limits(urdf_joint: urdf.Joint) -> Tuple[DerivativeMap[float], DerivativeMap[float]]:
+def urdf_joint_to_limits(urdf_joint: urdfpy.Joint) -> Tuple[DerivativeMap[float], DerivativeMap[float]]:
```
LucaKro, line 24: docstringm and maybe better Return Type. Maybe a class "JointLimits"?

src/semantic_world/adapters/urdf.py
```diff
@@ -102,15 +110,16 @@ def parse(self) -> World:
...
+                parsed_joint = self.parse_joint(joint, parent, child, world, prefix)
                 joints.append(parsed_joint)
 
             [world.add_connection(joint) for joint in joints]
             [world.add_body(link) for link in links]
 
         return world
 
-    def parse_joint(self, joint: urdf.Joint, parent: Body, child: Body, world: World) -> Connection:
+    def parse_joint(self, joint: urdfpy.Joint, parent: Body, child: Body, world: World, prefix: str) -> Connection:
```
LucaKro, line 121: docstring

src/semantic_world/geometry.py
```diff
@@ -91,6 +112,18 @@ def mesh(self) -> trimesh.Trimesh:
         """
         raise NotImplementedError("Subclasses must implement the mesh property.")
 
+    def is_bigger(self, volume_threshold: float = 1.001e-6, surface_threshold: float = 0.00061) -> bool:
```
LucaKro, line 115: huh?

src/semantic_world/geometry.py
```diff
@@ -183,6 +216,9 @@ def as_bounding_box(self) -> BoundingBox:
         return BoundingBox(-self.radius, -self.radius, -self.radius,
                            self.radius, self.radius, self.radius)
 
+    def is_bigger(self, volume_threshold: float = 1.001e-6, surface_threshold: float = 0.00061) -> bool:
```
LucaKro, line 219: i dont like this "is bigger". I feel like we should just have  property "volume" maybe, which you compare outside of the class?

src/semantic_world/geometry.py
```diff
@@ -329,7 +371,9 @@ def contains(self, point: Point3) -> bool:
         """
         Check if the bounding box contains a point.
         """
-        x, y, z = (point.x.to_np(), point.y.to_np(), point.z.to_np()) if isinstance(point.z, Expression) else (point.x, point.y, point.z)
+        x, y, z = (point.x.to_np(), point.y.to_np(), point.z.to_np()) if isinstance(point.z, Expression) else (point.x,
```
LucaKro, line 374: why do we need this expression check? I feel like this doesnt make sense for cases where to_np() doesnt work

src/semantic_world/robots.py
```diff
@@ -282,6 +286,16 @@ class AbstractRobot(RootedView, ABC):
     - an optional collection of sensor chains, each containing a sensor, such as a camera
     => If a kinematic chain contains both a manipulator and a sensor, it will be part of both collections
     """
+    odom: Body = field(default=None)
```
LucaKro, line 289: where in the kinematic chain is odom here?

src/semantic_world/robots.py
```diff
@@ -282,6 +286,16 @@ class AbstractRobot(RootedView, ABC):
...
+    """
+    The odometry body of the robot, which is usually the base footprint.
+    """
+
+    drive: Optional[OmniDrive] = None
```
LucaKro, line 294: why is OmniDrive the Type here? shouldnt it just be Drive,  where omnidrive, diffdrive etc inherit from?

src/semantic_world/robots.py
```diff
@@ -436,6 +471,8 @@ def from_world(cls, world: World) -> Self:
             _world=world,
         )
 
+        robot.drive = robot.root.parent_connection
```
LucaKro, line 474: the root doesnt have a parent connection. the parent connection is set as soon as we merge a world, otherwise it doesnt really make sense i think. now if odom was the root then i think it would make sense to add the drive as a connection between odom and the robots urdf root

src/semantic_world/spatial_types/math.py
```diff
@@ -326,3 +326,57 @@ def quaternion_slerp(q1, q2, t):
         return 0.5 * q1 + 0.5 * q2
     return ratio_a * q1 + ratio_b * q2
 
+
+
+def sphere_volume(radius):
```
LucaKro, line 331: this belings into the respective shape class as a property imo, same with the others below

src/semantic_world/adapters/urdf.py
```diff
@@ -64,11 +64,17 @@ def urdf_joint_to_limits(urdf_joint: urdf.Joint) -> Tuple[DerivativeMap[float],
...
     """
     Class to parse URDF files to worlds.
+    Must set either urdf or file_path.
     """
 
-    file_path: str
+    urdf: str = field(default=None, kw_only=True)
     """
-    The file path of the URDF.
+    The URDF string.
```
LucaKro, line 72: I think this class should have class methods "from_file_path()" and "from_urdf_string()" which instantiate this class, instead of this postinit checking

src/semantic_world/world_entity.py
```diff
@@ -222,12 +294,12 @@ def bounding_box_collection(self) -> BoundingBoxCollection:
         return BoundingBoxCollection(world_bboxes)
 
     @property
-    def global_pose(self) -> NpMatrix4x4:
+    def global_pose(self) -> cas.TransformationMatrix:
```
LucaKro, line 297: Isnt the result our transformation matrix?

PR #52, opened 2025-08-13
[robots] small robot view checkup

LucaKro:
- changed default values to None where proper,
- moved robot tests to new text folder test_robots, 
- fixed missing tip links,
- added "get_views_by_type" filtering method for worlds


src/semantic_world/robots.py
```diff
@@ -459,11 +459,13 @@ def from_world(cls, world: World) -> Self:
...
                                        finger=left_gripper_finger,
                                        _world=world)
-
-        left_arm = Arm(name=PrefixedName('left_arm', prefix=robot.name.name),
-                       root=world.get_body_by_name("l_shoulder_pan_link"),
-                       manipulator=left_gripper,
-                       _world=world)
+        left_arm = Arm(
+            name=PrefixedName("left_arm", prefix=robot.name.name),
+            root=world.get_body_by_name("l_shoulder_pan_link"),
```
Tigul, line 464: The root link of the arms is "torso_lift_link" because the l_shoulder_pan_joint is before the link

src/semantic_world/robots.py
```diff
@@ -484,6 +486,7 @@ def from_world(cls, world: World) -> Self:
                                         _world=world)
         right_arm = Arm(name=PrefixedName('right_arm', prefix=robot.name.name),
                         root=world.get_body_by_name("r_shoulder_pan_link"),
```
Tigul, line 488: Same here

Tigul:
In the world.add_view(view) please add exist_ok=True

LucaKro:
> In the world.add_view(view) please add exist_ok=True

tbh I dont get why this is even a thing, I think we should just have world.views be a set, you should always query the world for specific views, not access the world.views property directly, so it being unsorted doesnt matter

Tigul:
exist_ok=True doesn"t raise an error if the view already exists

Tigul (approved):

tomsch420 (approved):

PR #55, opened 2025-08-22
[View Factories] Factories for creating views based on input parameters

LucaKro:
# Feature: ViewFactories

With this PR @tomsch420 and I introduce ViewFactories to the semantic digital twin. This enables us to create our own articulated furniture models. We can influence the specifics of the created body collisions using input parameters.

image

We can then fill the interior of the body using random events, so that we are left with a solid Dresser with 4 Drawers and 4 Doors.

The input parameters were chosen based on the information we are provided by procthor json files as well as fbx files, as this will be our first usecase for these factories in a future PR.

---

## Other Changes

- the as_bounding_box methods of shapes was generalized and pulled up into the shape super class. The variable names were adjusted to adhere to the style guide
- orphaned DoF are now removed from a world during model change. in that regard we also now have 'register_degree_of_freedom' method, since we may want to create DoF by hand at some points, and doing that using the 'create_degree_of_freedom' without being in another 'with world.modify_world()' context block adds a DoF to the world, notifies the model change, and then instantly deletes it again. So now we have a explicit method for registering the dof.
- Merge_world now correctly also transfers the views to the new world
- Adds xz and yz spatial variable combinations to SpatialVariables enum

---

Credit: Layout inspired by [SciPy](https://docs.scipy.org/doc/scipy/release/1.15.0-notes.html)

### DIFF
--- a/src/semantic_world/geometry.py
+++ b/src/semantic_world/geometry.py
@@ -1,12 +1,15 @@
 from __future__ import annotations
 
 import itertools
+import logging
 import tempfile
-from abc import ABC
+from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from functools import cached_property
-from typing import Optional, List, Iterator
+from typing import Optional, List, Iterator, TYPE_CHECKING
 
+import numpy as np
+from numpy import ndarray
 from trimesh import Trimesh
 from typing_extensions import Self
 
@@ -14,10 +17,16 @@ import trimesh
 from random_events.interval import SimpleInterval, Bound
 from random_events.product_algebra import SimpleEvent, Event
 
+from .prefixed_name import PrefixedName
 from .spatial_types import TransformationMatrix, Point3
 from .spatial_types.spatial_types import Expression
+from .types import NpMatrix4x4
 from .utils import IDGenerator
 from .variables import SpatialVariables
+import trimesh.exchange.stl
+
+if TYPE_CHECKING:
+    from .world_entity import Body
 
 id_generator = IDGenerator()
 
@@ -76,12 +85,49 @@ class Shape(ABC):
     """
     origin: TransformationMatrix = field(default_factory=TransformationMatrix)
 
-    def as_bounding_box(self) -> BoundingBox:
+    def as_bounding_box(self, reference_frame: Optional[Body]) -> BoundingBox:
         """
-        Returns the bounding box of the shape.
+        Returns the bounding box of the box.
+        """
+
+        origin_frame = self.origin.reference_frame
+        origin_T_self = self.origin
+
+        if not reference_frame:
+            reference_frame = self.origin.reference_frame
+
+        reference_frame_world = reference_frame._world or origin_frame._world
+
+        if reference_frame_world is not None:
+            reference_T_origin: TransformationMatrix = reference_frame_world.compute_forward_kinematics(reference_frame,
+                                                                                           origin_frame)
+        else:
+            reference_T_origin: TransformationMatrix = TransformationMatrix()
+
+        reference_T_self: TransformationMatrix = reference_T_origin @ origin_T_self
+
+        # Get all 8 corners of the BB in link-local space
+        list_self_T_corner = [TransformationMatrix.from_point_rotation_matrix(self_P_corner) for self_P_corner in self._local_bounding_box().get_points()] # shape (8, 3)
+
+        list_reference_T_corner = [reference_T_self @ self_T_corner for self_T_corner in list_self_T_corner]
+
+        list_reference_P_corner = [reference_T_corner.to_position().to_np()[:3] for reference_T_corner in list_reference_T_corner]
+
+        # Compute world-space bounding box from transformed corners
+        min_corner = np.min(list_reference_P_corner, axis=0)
+        max_corner = np.max(list_reference_P_corner, axis=0)
+
+        world_bb = BoundingBox.from_min_max(Point3.from_iterable(min_corner), Point3.from_iterable(max_corner))
+
+        return world_bb
+
+    @abstractmethod
+    def _local_bounding_box(self) -> BoundingBox:
+        """
+        Returns the local bounding box of the shape.
         This method should be implemented by subclasses.
         """
-        raise NotImplementedError("Subclasses must implement this method.")
+        ...
 
     @property
     def mesh(self) -> trimesh.Trimesh:
@@ -112,14 +158,15 @@ class Mesh(Shape):
         """
         The mesh object.
         """
-        return trimesh.load_mesh(self.filename)
+        mesh = trimesh.load_mesh(self.filename)
+        return mesh
 
-    def as_bounding_box(self) -> BoundingBox:
+    def _local_bounding_box(self) -> BoundingBox:
         """
-        Returns the bounding box of the mesh.
+        Returns the local bounding box of the mesh.
+        The bounding box is axis-aligned and centered at the origin.
         """
         return BoundingBox.from_mesh(self.mesh)
-
 
 @dataclass
 class TriangleMesh(Shape):
@@ -148,10 +195,8 @@ class TriangleMesh(Shape):
         return BoundingBox.from_mesh(self.mesh)
 
 
-
-
 @dataclass
-class Primitive(Shape):
+class Primitive(Shape, ABC):
     """
     A primitive shape.
     """
@@ -176,7 +221,7 @@ class Sphere(Primitive):
         """
         return trimesh.creation.icosphere(subdivisions=2, radius=self.radius)
 
-    def as_bounding_box(self) -> BoundingBox:
+    def _local_bounding_box(self) -> BoundingBox:
         """
         Returns the bounding box of the sphere.
         """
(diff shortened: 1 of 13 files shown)


ichumuh:
> @ichumuh can we add you to the project so that i can add you as a reviewer? :D

sure, add me

LucaKro:
> > @ichumuh can we add you to the project so that i can add you as a reviewer? :D
> 
> sure, add me

@Tigul can you add him? I dont think I have the perms

src/semantic_world/views/views.py
```diff
@@ -98,7 +125,18 @@ class Door(Components):
...
     def __post_init__(self):
-        self.name = PrefixedName(str(self.body.name), self.__class__.__name__)
+        if self.name is None:
+            self.name = PrefixedName(str(self.body.name), self.__class__.__name__)
+
+
+@dataclass(unsafe_hash=True)
+class DoubleDoor(Components):
+    body: Body
+    doors: List[Door] = field(default_factory=list, hash=False)
```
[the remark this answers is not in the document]
tomsch420: i see and agree.

tomsch420 (approved):

src/semantic_world/world_entity.py
```diff
@@ -170,53 +171,44 @@ def evaluated_geometric_distribution(n: int) -> np.ndarray:
...
         """
         return self._world.compute_child_bodies(self)
 
+    @property
+    def recursive_child_bodies(self) -> List[Body]:
+        """
+        Returns the recursive child bodies of this body.
+        :return: All child bodies of this body.
+        """
+        return [self._world.kinematic_structure[i] for i in rustworkx.descendants(self._world.kinematic_structure, self.index)]
```
ichumuh, line 184: That's not a good name. recursive implies a method to find child bodies, but 1. the method doesn't matter and shouldn't be part of the name and 2. the method isn't even recursive.
name it decendant_child_bodies or something

src/semantic_world/world.py
```diff
@@ -362,11 +386,18 @@ def _notify_model_change(self) -> None:
             # self._fix_tree_structure()
             self.reset_cache()
             self.compile_forward_kinematics_expressions()
-            # self._cleanup_unused_dofs()
+            self.deleted_orphaned_dof()
             self.notify_state_change()
             self._model_version += 1
             self.validate()
 
+    def deleted_orphaned_dof(self):
```
ichumuh, line 394: `delete_orphaned_dofs`

src/semantic_world/world_entity.py
```diff
@@ -170,53 +171,44 @@ def evaluated_geometric_distribution(n: int) -> np.ndarray:
...
-        Note: These bounding boxes may not be disjoint, however the random events library always makes them disjoint. If
-        this is the case, and we feed he non-disjoint bounding boxes into the gcs, it may trigger unexpected behavior.
+        This property computes the forward kinematics to determine the object's
+        position and orientation in world space. Bounding boxes for each shape in the
+        collision attribute are transformed to world coordinates. The world-space bounding
+        boxes are then aggregated into a BoundingBoxCollection.
 
-        :return: A BoundingBoxCollection containing the bounding boxes of the link's geometry in world
+        :returns: A collection of bounding boxes in world-space coordinates.
+        :rtype: BoundingBoxCollection
```
ichumuh, line 204: update doc string.
why is reference frame optional?

src/semantic_world/world_entity.py
```diff
@@ -293,13 +284,13 @@ def bodies(self) -> Iterable[Body]:
         """
         return self._bodies(set())
 
-    def as_bounding_box_collection(self) -> BoundingBoxCollection:
+    def as_bounding_box_collection(self, reference_frame: Optional[Body] = None) -> BoundingBoxCollection:
```
ichumuh, line 287: why is reference frame optional?

src/semantic_world/world_entity.py
```diff
@@ -293,13 +284,13 @@ def bodies(self) -> Iterable[Body]:
...
         """
         Returns a bounding box collection that contains the bounding boxes of all bodies in this view.
         """
         bbs = reduce(
             lambda accumulator, bb_collection: accumulator.merge(bb_collection),
-            (body.bounding_box_collection for body in self.bodies if body.has_collision())
+            (body.as_bounding_box_collection(reference_frame) for body in self.bodies if body.has_collision())
```
ichumuh, line 293: sorry, but this is a horribly complicated line of code. this should be a loop instead.

src/semantic_world/geometry.py
```diff
@@ -76,12 +85,53 @@ class Shape(ABC):
     """
     origin: TransformationMatrix = field(default_factory=TransformationMatrix)
 
-    def as_bounding_box(self) -> BoundingBox:
+    def as_bounding_box(self, reference_frame: Optional[Body]) -> BoundingBox:
```
ichumuh, line 88: it's feels weird that reference frame can be optional. you are computing it using information from this shape, so it should always have the same reference frame as the shape, no?
btw, even if this stays optional, it should have default None, otherwise its not really optional

src/semantic_world/geometry.py
```diff
@@ -76,12 +85,53 @@ class Shape(ABC):
...
+        max_corner = np.max(list_reference_P_corner, axis=0)
+
+        world_bb = BoundingBox.from_min_max(Point3.from_iterable(min_corner), Point3.from_iterable(max_corner))
+
+        return world_bb
+
+    @abstractmethod
+    def _local_bounding_box(self) -> BoundingBox:
+        """
+        Returns the local bounding box of the shape.
```
ichumuh, line 131: what's a local bounding box?

src/semantic_world/geometry.py
```diff
@@ -427,6 +485,15 @@ def from_min_max(cls, min_point: Point3, max_point: Point3) -> Self:
...
 
+    def as_shape(self, reference_frame: Optional[Body] = None) -> Box:
+        scale = Scale(x=self.max_x - self.min_x,
+                      y=self.max_y - self.min_y,
+                      z=self.max_z - self.min_z)
+        x = (self.max_x + self.min_x) / 2
+        y = (self.max_y + self.min_y) / 2
+        z = (self.max_z + self.min_z) / 2
+        origin = TransformationMatrix.from_xyz_rpy(x, y, z, 0, 0, 0, reference_frame)
+        return Box(origin=origin, scale=scale)
```
ichumuh, line 496: it's very weird that a shape has an origin but a bounding box doesn't

src/semantic_world/views/views.py
```diff
@@ -1,50 +1,76 @@
...
+
     handle: Handle
     body: Body
 
     def __post_init__(self):
         self.name = PrefixedName(str(self.body.name), self.__class__.__name__)
 
+
 @dataclass(unsafe_hash=True)
 class Fridge(View):
```
ichumuh, line 66: shouldn't a fridge inherit from `HasDoor` and likely also `HasDrawers`?

src/semantic_world/views/views.py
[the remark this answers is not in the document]
ichumuh: but now doubledoors can have more or less than two doors. even if left/right is bad, i still think it should be two explicit doors, maybe door1 and door2

src/semantic_world/geometry.py
LucaKro: boundingbox from its own frame of reference, imma update the docstring

src/semantic_world/geometry.py
ichumuh: can you give me an example?

src/semantic_world/geometry.py
ichumuh: doesn't that mean the implementation should be `return self.as_bounding_box(reference_frame=None)`?

src/semantic_world/geometry.py
ichumuh: hm idk, doesn't that violate the single responsibility principle?

src/semantic_world/geometry.py
LucaKro: nope, since as_bounding_box uses _local_bounding_box.

local bounding box is the method all shapes need to override, since somewhere how to get the bounding box for this shape needs to be defined. This allows as_bounding_box to be defined in a generalised way.

src/semantic_world/world.py
```diff
@@ -527,16 +558,23 @@ def merge_world(self, other: World, root_connection: Connection = None) -> None:
...
-
         for body in other.bodies:
-            other.remove_body(body)
-            self.add_body(body)
-        other.world_is_being_modified = False
+            if body._world is not None:
+                other.remove_body(body)
 
+        for view in other.views:
+            self.add_view(view, exists_ok=True)
```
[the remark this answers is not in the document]
LucaKro: I am really not a fan of exists_ok. 
1. I dont think having the same view twice in the system warrants an error, its a bit inefficient at worst, but not a invalid state.
2. If we really want to avoid that we should just use sets, the user shouldnt access world.views directly anyways, they can always just query by name or type.
3. and if there can actually be a case where a duplicate view would create an invalid state, then I think we shouldnt provide an exists_ok and instead just force the user to handle the views correctly

src/semantic_world/world_entity.py
LucaKro: it was not used anywhere, so i removed it for now. If i need to add it again later ill rename it

PR #56, opened 2025-08-25
[view factory] PR Touchup 

LucaKro:
There was a review in the works at the time https://github.com/cram2/semantic_world/pull/55 got merged. This PR addresses the feedback from that review

### DIFF
--- a/src/semantic_world/geometry.py
+++ b/src/semantic_world/geometry.py
@@ -1,29 +1,24 @@
 from __future__ import annotations
 
 import itertools
-import logging
 import tempfile
-from abc import ABC, abstractmethod
+from abc import ABC, abstractmethod, abstractproperty
 from dataclasses import dataclass, field
 from functools import cached_property
 from typing import Optional, List, Iterator, TYPE_CHECKING
 
 import numpy as np
-from numpy import ndarray
+import trimesh
+import trimesh.exchange.stl
+from random_events.interval import SimpleInterval, Bound
+from random_events.product_algebra import SimpleEvent, Event
 from trimesh import Trimesh
 from typing_extensions import Self
 
-import trimesh
-from random_events.interval import SimpleInterval, Bound
-from random_events.product_algebra import SimpleEvent, Event
-
-from .prefixed_name import PrefixedName
 from .spatial_types import TransformationMatrix, Point3
 from .spatial_types.spatial_types import Expression
-from .types import NpMatrix4x4
 from .utils import IDGenerator
 from .variables import SpatialVariables
-import trimesh.exchange.stl
 
 if TYPE_CHECKING:
     from .world_entity import Body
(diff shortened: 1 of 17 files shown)


src/semantic_world/geometry.py
```diff
@@ -476,33 +456,88 @@ def get_points(self) -> List[Point3]:
                 for z in (self.min_z, self.max_z)]
 
     @classmethod
-    def from_min_max(cls, min_point: Point3, max_point: Point3) -> Self:
+    def from_min_max(
+        cls, reference_frame: Body, min_point: Point3, max_point: Point3
+    ) -> Self:
```
ichumuh, line 461: Point3 has a reference frame. so you should probably not have it as a parameter in from_min_max, but verify that the two points have the same reference frame and the use it.

src/semantic_world/geometry.py
```diff
@@ -476,33 +456,88 @@ def get_points(self) -> List[Point3]:
...
 class BoundingBoxCollection:
     """
     Dataclass for storing a collection of bounding boxes.
     """
 
+    reference_frame: Body
+    """
+    The reference frame of the bounding boxes.
+    """
+
```
ichumuh, line 536: i don't think a bounding box collection needs a reference frame. 
the bounding boxes already have one, so what would the frame of a collection mean?

ichumuh: if it's the frame that all bounding boxes should have, then make sure in __post_init__ that they do and give the collection a property instead, if you need.

src/semantic_world/world.py
```diff
@@ -299,39 +297,30 @@ def validate(self) -> bool:
...
-    def register_degree_of_freedom(self, dof: DegreeOfFreedom) -> None:
+    def add_degree_of_freedom(self, dof: DegreeOfFreedom) -> None:
         """
         Register a degree of freedom in the world.
         This is used to register DoFs that are not created by the world, but are part of the world model.
         :param dof: The degree of freedom to register.
         """
-        if dof in self.degrees_of_freedom:
+        if dof._world is self:
             logger.debug(f"Degree of freedom {dof.name} already registered in world {self.name}.")
```
ichumuh, line 316: still not a fan of this debug message. also, if you set the `_world` of a degree of freedom manually for some reason, it would silently not get added.

src/semantic_world/world.py
```diff
@@ -299,39 +297,30 @@ def validate(self) -> bool:
...
             return
+
+        if dof._world is not None:
+            raise NotImplementedError("Cannot register a degree of freedom that already belongs to another world.")
```
ichumuh, line 320: NotImplementedError is not the correct exception

ichumuh: maybe valueerror or some custom thing

src/semantic_world/world.py
ichumuh: this doesn't solve the issue. if you don't want exceptions that at least to `dof._world is self and dof in self.degrees_of_freedom`

LucaKro: done

ichumuh (approved):

PR #63, opened 2025-09-01
Re-added get -and add body

Tigul:
Re-adds the method get_body_by_name and add_body for better use with typing and linting.

### DIFF
--- a/src/semantic_world/world.py
+++ b/src/semantic_world/world.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 from __future__ import annotations
 
+import inspect
 import logging
 import os
 from copy import deepcopy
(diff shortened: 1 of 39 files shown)


src/semantic_world/world.py
```diff
@@ -503,6 +503,9 @@ def add_kinematic_structure_entity(
         )
         kinematic_structure_entity._world = self
 
+    def add_body(self, body: Body) -> None:
```
LucaKro, line 506: get_body_by_name makes sense, but add_body is unnecessary i think

LucaKro (changes requested):

src/semantic_world/world.py
LucaKro: I really dont think its necessary. If that is really your position, then it should include checks that its actually a body I think.

right now you can do `world.add_body(region)`, without any problems, which i dont like.

tbh I feel like add_body is being added to get around the naming discussion we had earlier :D

Tigul: No that sound like a weired to use interface

tomsch420: in a pure definition world you would be right @LucaKro , i just find it inconvenient and agree that this changes is a good QOL change.

PR #64, opened 2025-09-01
Integrate my ontology changes into views

mkhoshnam:
This PR introduces modifications to src/semantic_world/views/views.py to extend the ontology support.
Specifically:

Added new ontology-related fields to OntologyView (e.g., affordances, manipulation_properties, state_information, grasp_poses, effects, description, joint_type, joint_sim_name).

Updated relevant classes (Drawer, Cabinet, Fridge, Microwave, Oven, Stove, Sink, Dishwasher, Counter, CoffeeMachine, Hood, etc.) to inherit from or integrate with OntologyView.

Refactored portions of the file to support richer semantic information in the views layer.

Purpose

These changes allow integration of a custom ontology with the semantic_world framework, enabling representation of affordances, manipulation properties, and semantic states for objects like drawers, cabinets, fridges, microwaves, etc.

This will make the ontology more expressive and easier to use in downstream reasoning and planning tasks.

Purpose

These changes allow integration of a custom ontology with the semantic_world framework, enabling representation of affordances, manipulation properties, and semantic states for objects like drawers, cabinets, fridges, microwaves, etc.

This will make the ontology more expressive and easier to use in downstream reasoning and planning tasks.

### DIFF
--- a/src/semantic_world/adapters/digital_twin_adapter.py
+++ b/src/semantic_world/adapters/digital_twin_adapter.py
@@ -7,6 +7,8 @@ Adjust the imports below to match your repo's layout.
 
 from dataclasses import dataclass
 from typing import Optional
+import inspect
+from types import SimpleNamespace
 
 import mujoco as mj
 
@@ -14,6 +16,41 @@ import mujoco as mj
 from semantic_world.world import Body
 from semantic_world.datastructures.prefixed_name import PrefixedName
 from semantic_world.views.views import Cabinet, Door, Handle  # your dataclasses
+
+# ---------- Flexible constructor helper ----------
+
+def _flex_construct(cls, fallback_attrs: dict):
+    """
+    Try to instantiate `cls` by filtering kwargs to what its __init__ accepts.
+    Then, ensure all attrs in fallback_attrs exist on the instance (setattr if missing).
+    If construction fails, return a SimpleNamespace with those attrs.
+    """
+    # Figure out the signature (dataclasses may expose __init__, sometimes cls itself)
+    try:
+        sig = inspect.signature(cls.__init__)
+    except (TypeError, ValueError, AttributeError):
+        try:
+            sig = inspect.signature(cls)
+        except Exception:
+            sig = None
+
+    kwargs = {}
+    if sig is not None:
+        for k, v in fallback_attrs.items():
+            if k in sig.parameters:
+                kwargs[k] = v
+
+    try:
+        obj = cls(**kwargs)
+    except Exception:
+        obj = SimpleNamespace(**fallback_attrs)
+        return obj
+
+    # Backfill any attrs we rely on but that weren't constructor params
+    for k, v in fallback_attrs.items():
+        if not hasattr(obj, k):
+            setattr(obj, k, v)
+    return obj
 
 # ---------- Build a semantic Cabinet view from a MuJoCo model ----------
 
@@ -49,8 +86,8 @@ def build_cabinet_view(
     d_body   = _mk_body(model, data, door_body)
     h_body   = _mk_body(model, data, handle_body)
 
-    handle = Handle(body=h_body)
-    door   = Door(body=d_body, handle=handle)
+    handle = _flex_construct(Handle, {"body": h_body})
+    door   = _flex_construct(Door,   {"body": d_body, "handle": handle})
 
     # Derive joint type if provided
     joint_type = None
@@ -65,14 +102,17 @@ def build_cabinet_view(
             else:
                 joint_type = "other"
 
-    return Cabinet(
-        body=cab_body,
-        doors=[door],
-        drawers=[],
-        joint_type=joint_type,
-        joint_sim_name=door_joint,
-        description="Semantic view built from MuJoCo model",
-    )
+    cab_fallback = {
+        "body": cab_body,
+        "doors": [door],
+        "drawers": [],
+        "joint_type": joint_type,
+        "joint_sim_name": door_joint,
+        "description": "Semantic view built from MuJoCo model",
+    }
+
+    cab = _flex_construct(Cabinet, cab_fallback)
+    return cab
 
 # ---------- Digital-Twin Controller bound to the semantic view ----------
 
@@ -138,4 +178,3 @@ class DigitalTwinCabinetAgent:
             return val > 0.25 * (thresh_percent / 100.0)
         else:
             return False
-


src/semantic_world/views/views.py
```diff
@@ -13,6 +13,21 @@
 from ..world import View, Body
 
 
+@dataclass
+class OntologyView(View):
```
AbdelrhmanBassiouny, line 17: I think OntologyView doesn't convey what this is about, also I think this class shouldn't be a view since it has no relation to the world by itself.

from how this class looks I think something like `ManipulationInfo` would make more sense, and it could be that it's better to use composition here instead of inheritance like this:
```python
@dataclass
class Drawer(View):
    manipulation_info: ManipulationInfo = field(default_factory=ManipulationInfo)
```

AbdelrhmanBassiouny (changes requested):

mkhoshnam:
Implemented requested changes: replaced OntologyView with ManipulationInfo , updated all relevant views to hold manipulation_info, and added per-field documentation in the ManipulationInfo docstring. Ready for re-review.

tomsch420:
Also make the class more specific. If  it  handles information like this in general, make it an abstract class and create subclasses for specific kinds of ManipulationInformation like this. I think you would benefit from a use case and not just the addition of random fields to all views. Perhaps come see us in the office to discuss what you want to do.

tomsch420:
I think we are talking past each other here. Come to our office tomorrow.

src/semantic_world/adapters/digital_twin_adapter.py
```diff
@@ -0,0 +1,180 @@
...
+    Build a lightweight Body compatible with your repo's dataclass.
+    We avoid passing model/data/id because your Body(...) doesn't accept them.
+    """
+    # Optional sanity: ensure the body exists in the MuJoCo model
+    bid = mj.mj_name2id(model, mj.mjtObj.mjOBJ_BODY, body_name)
+    if bid < 0:
+        raise ValueError(f"Body '{body_name}' not found in MuJoCo model")
+
+    # Prefer PrefixedName if your Body expects it; fall back to plain str if needed.
+    try:
```
AbdelrhmanBassiouny, line 68: the Body takes a prefixed name no need for this try ... except

src/semantic_world/adapters/digital_twin_adapter.py
```diff
@@ -0,0 +1,180 @@
...
+        "description": "Semantic view built from MuJoCo model",
+    }
+
+    cab = _flex_construct(Cabinet, cab_fallback)
+    return cab
+
+# ---------- Digital-Twin Controller bound to the semantic view ----------
+
+@dataclass
+class DigitalTwinCabinetAgent:
```
AbdelrhmanBassiouny, line 120: It's very wierd to make a class that is considered an Agent that only controls the cabinet ... in addition sematnic world has APIs that you can use to manipulate the world state and do forward and inverse kinematics, please check the user start guide that was just merged for how to do that.

AbdelrhmanBassiouny (changes requested):
I do not think that this is useful to you it seems like a forced integration of semantic world to your work that does not seem to add any meaningful benefits to semantic world or to your work.

To benefit from semantic world you have to use it's features, for example you are trying to control a cabinet, and you did not use semantic world state or joints or degrees of freedom or the ik solver.... instead you are just using mujoco.

So please check the start user guide to be able to make use of semantic world.

Second thing the way you are currently constructing views from ontology/mujoco data were you hope you find matching attributes and if not just set them manually is not a good idea. It would be better if you think about your task and what are the needed attributes that would add a real value to semantic world views and then add them manually to the views because then it becomes useful to everyone not just your application.

src/semantic_world/adapters/digital_twin_adapter.py
LucaKro, line 1: why "digital_twin_adapter"?

src/semantic_world/adapters/digital_twin_adapter.py
```diff
@@ -0,0 +1,180 @@
+"""
+Digital-Twin adaptor: bridges semantic_world views to an actual simulator (MuJoCo).
+Use case: open / close a cabinet door by name, using the joint mapped from the semantic view.
+
+Adjust the imports below to match your repo's layout.
```
LucaKro, line 5: this also seems copy pasted from an LLM, I dont think we need this here

src/semantic_world/adapters/digital_twin_adapter.py
```diff
@@ -0,0 +1,180 @@
...
+
+    # Backfill any attrs we rely on but that weren't constructor params
+    for k, v in fallback_attrs.items():
+        if not hasattr(obj, k):
+            setattr(obj, k, v)
+    return obj
+
+# ---------- Build a semantic Cabinet view from a MuJoCo model ----------
+
+def _mk_body(model: mj.MjModel, data: mj.MjData, body_name: str) -> Body:
```
LucaKro, line 57: why do we have this? this method basically creates a body if the name is already in the mujoco model? but the object you create doesnt have any information about collision/visuals, and is also not added to the kinematic structure. trying to control a cabinet where none of the bodies are in the kinematic structure does not make any sense

src/semantic_world/adapters/digital_twin_adapter.py
```diff
@@ -0,0 +1,180 @@
...
+    door_body: str,
+    handle_body: str,
+    door_joint: Optional[str] = None
+) -> Cabinet:
+    """Create a Cabinet→Door→Handle semantic view from MuJoCo names."""
+    cab_body = _mk_body(model, data, cabinet_body)
+    d_body   = _mk_body(model, data, door_body)
+    h_body   = _mk_body(model, data, handle_body)
+
+    handle = _flex_construct(Handle, {"body": h_body})
```
LucaKro, line 89: why do we need "_flex_construct"? why cant we just do `Handle(body=h_body)` and `Door(body=d_body, handle=handle)`?

src/semantic_world/adapters/digital_twin_adapter.py
```diff
@@ -0,0 +1,180 @@
...
+            elif jtype == mj.mjtJoint.mjJNT_SLIDE:
+                joint_type = "prismatic"
+            else:
+                joint_type = "other"
+
+    cab_fallback = {
+        "body": cab_body,
+        "doors": [door],
+        "drawers": [],
+        "joint_type": joint_type,
```
LucaKro, line 109: we already have a concept of joints in the semdt, why are you doing it like this instead of creating a connection of the proper kind?

src/semantic_world/adapters/digital_twin_adapter.py
```diff
@@ -0,0 +1,180 @@
...
+import mujoco as mj
+
+# ⬇️ Adjust these imports if your module paths differ
+from semantic_world.world import Body
+from semantic_world.datastructures.prefixed_name import PrefixedName
+from semantic_world.views.views import Cabinet, Door, Handle  # your dataclasses
+
+# ---------- Flexible constructor helper ----------
+
+def _flex_construct(cls, fallback_attrs: dict):
```
LucaKro, line 22: i dont like this method exists, I dont think its necessary.

also the name"fallback_attrs" doesnt make any sense to me here

src/semantic_world/adapters/digital_twin_adapter.py
```diff
@@ -0,0 +1,180 @@
...
+        jid = mj.mj_name2id(self.model, mj.mjtObj.mjOBJ_JOINT, self.cab.joint_sim_name)
+        if jid < 0:
+            raise RuntimeError(f"Joint '{self.cab.joint_sim_name}' not found in model.")
+        return jid
+
+    def set_fraction_open(self, frac: float) -> None:
+        """
+        Set door openness in [0,1].
+        - hinge: maps to angle between [closed, open], here we assume [0, +90°]
+        - prismatic: maps to linear slide, assume [0, +0.25 m]
```
LucaKro, line 139: why do we "assume"? doesnt mujoco joints have jointlimits you can use here?

src/semantic_world/adapters/digital_twin_adapter.py
LucaKro: I also dont get how this is an agent yet, but i may be misunderstanding

LucaKro (changes requested):
while this looks very interesting, I dont think the content of this PR fit with the current vision of semdt yet. to me this looks to me more like the start of its own framework that imports semdt, similar to robokudo, giskard, multiverse etc.
So if i were you I would create my own package that you implement this functionality inside, and import semdt for features you need that we have already implemented. if you notice that some feature is missing in semdt, then you create a PR adding that feature you need.

since you are also working with mujoco there may be some crossover with multiverse, but im not sure, you would need to talk to giang about that.

PR #77, opened 2025-09-15
Shape collection

tomsch420:
The type List[Shape] has been replaced by ShapeCollection since some classes share methods for list of shapes. This reduced code duplication.


src/semantic_world/views/views.py
```diff
@@ -37,20 +37,12 @@ class HasDoors:
 class Handle(View):
     body: Body
 
-    def __post_init__(self):
```
LucaKro, line 40: i think there was a reason we had names for these if they were not provided

src/semantic_world/world_description/geometry.py
```diff
@@ -23,7 +22,7 @@
 from ..utils import IDGenerator
 
 if TYPE_CHECKING:
-    from ..world_description.world_entity import KinematicStructureEntity
+    pass
```
LucaKro, line 25: remove this then?

src/semantic_world/world_description/shape_collection.py
```diff
@@ -0,0 +1,283 @@
...
+
+    reference_frame: Optional[KinematicStructureEntity] = None
+    """
+    Backreference to the kinematic structure entity this collection belongs to.
+    """
+
+    def __post_init__(self):
+        if self.reference_frame is not None:
+            for shape in self.shapes:
+                shape.origin.reference_frame = self.reference_frame
```
LucaKro, line 41: uhm i dont think this is good to do like this. if for some reason the shape provided had a different reference frame, this creates weird bugs.

I think you should check if the shape already has a reference frame, and if it already has one either transform it into the correct frame, or throw an error

src/semantic_world/world_description/shape_collection.py
```diff
@@ -0,0 +1,283 @@
...
+    ) -> BoundingBoxCollection:
+        """
+        Provides the bounding box collection for this entity given a transformation matrix as origin.
+        :param origin: The origin to express the bounding boxes from.
+        :returns: A collection of bounding boxes in world-space coordinates.
+        """
+        world_bboxes = []
+
+        for shape in self.shapes:
+            if shape.origin.reference_frame is None:
```
LucaKro, line 83: this case should never happen right? since you set this at post init.

that being said, maybe it makes sense to also check and replace the reference frame when appending

src/semantic_world/world_description/shape_collection.py
```diff
@@ -0,0 +1,283 @@
...
+                reference_frame,
+            )
+
+    def as_shapes(self) -> ShapeCollection:
+        return ShapeCollection(
+            [box.as_shape() for box in self.bounding_boxes],
+            self.reference_frame,
+        )
+
+    def get_points(self) -> List[Point3]:
```
LucaKro, line 263: this essentially creates a bounding box around the bounding box collection no?

src/semantic_world/world_description/world_entity.py
```diff
@@ -197,19 +175,20 @@ class Body(KinematicStructureEntity, SubclassJSONSerializer):
...
-        return concatenate(transformed_meshes)
+    def __post_init__(self):
+        if not self.name:
+            self.name = PrefixedName(f"body_{id_generator(self)}")
+
+        if self._world is not None:
+            self.index = self._world.kinematic_structure.add_node(self)
+
+        for c in self.collision:
+            c.origin.reference_frame = self
```
LucaKro, line 186: same issue here as in shapecollection. just overriding the reference frame without checking can cause really frustrating and hard to debug errors. maybe throw an error if the reference frame is set and is not self, or transform the shape into the correct frame

LucaKro (changes requested):
nice good work!

src/semantic_world/world_description/graph_of_convex_sets.py
```diff
@@ -240,15 +241,20 @@ def _make_search_space(
...
         if search_space is None:
-            search_space = BoundingBox(
-                min_x=-np.inf,
-                min_y=-np.inf,
-                min_z=-np.inf,
-                max_x=np.inf,
-                max_y=np.inf,
-                max_z=np.inf,
-                origin=TransformationMatrix(reference_frame=world.root),
-            ).as_collection()
```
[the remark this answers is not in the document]
tomsch420: That causes cyclic Imports :(

src/semantic_world/world_description/shape_collection.py
tomsch420: Idk, i copied it from the existing

src/semantic_world/views/views.py
tomsch420: The namens still exist but are handled by the superclass

src/semantic_world/world_description/shape_collection.py
tomsch420: this can happen if shape.origin frame is None and my frame is None.

src/semantic_world/world_description/shape_collection.py
tomsch420: so what do you suggest concretely? Return a bb? Change docstring?

src/semantic_world/world_description/shape_collection.py
```diff
@@ -53,6 +77,10 @@ def __contains__(self, shape: Shape) -> bool:
         return shape in self.shapes
 
     def append(self, shape: Shape):
+        # if self.reference_frame is not None:
```
LucaKro, line 80: did this not work? or why was this comment added?

tomsch420: not really, i will try soemthing else today but perhaps the user just has to manage the alignment of the frames

src/semantic_world/adapters/urdf.py
```diff
@@ -293,19 +296,10 @@ def parse_geometry(
...
-                else (
-                    0,
-                    0,
-                    0,
-                    0,
-                    0,
-                    0,
-                )
+            params = (*(geom.origin.xyz + geom.origin.rpy),) if geom.origin else ()
+            origin_transform = TransformationMatrix.from_xyz_rpy(
```
LucaKro, line 300: nice

src/semantic_world/world_description/shape_collection.py
LucaKro: i think we can uncomment this now right? i think it makes sense to do this when appending, and it seems to work above?

src/semantic_world/world_description/shape_collection.py
```diff
@@ -0,0 +1,320 @@
...
+        """
+        if shape.origin.reference_frame is None:
+            # If we don’t have a world, fall back to the owning body/frame
+            shape.origin.reference_frame = self.reference_frame
+        elif (
+            self.reference_frame is not None
+            and shape.origin.reference_frame != self.reference_frame
+            and self.reference_frame._world is not None
+        ):
+            shape.origin = self.reference_frame._world.transform(
```
LucaKro, line 70: oh maybe do a warning here? otherwise unexperienced users may run into issues without any notice because they added shapes in the wrong reference frame to the body, and suddenly have a completely wrong collision?

LucaKro (approved):
Nice! good job!

PR #89, opened 2025-09-24
World factory

LucaKro:
TBD

### DIFF
--- a/src/semantic_world/views/factories.py
+++ b/src/semantic_world/views/factories.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
+import logging
 from dataclasses import dataclass, field
-from enum import IntEnum
+from enum import IntEnum, Enum
 
 from numpy import ndarray
+from probabilistic_model.probabilistic_circuit.rx.helper import uniform_measure_of_event
+from random_events.interval import Bound
 from random_events.product_algebra import *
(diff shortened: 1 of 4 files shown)


src/semantic_world/views/factories.py
```diff
@@ -329,35 +390,75 @@ def _zoom_interval(base: SimpleInterval, target: SimpleInterval) -> SimpleInterv
...
-            {SpatialVariables.y.value: cur_y, SpatialVariables.z.value: cur_z}
-        )
+                case SimpleInterval():
+                    current_interval = self._zoom_interval(current_interval, step)
+                case DirectionDirac():
+                    current_dirac = self._zoom_dirac(current_interval, step)
+                    current_dirac = make_dirac(axis, float(current_dirac))
+                    break
+
+        return current_dirac or SimpleEvent({axis: current_interval})
```
tomsch420, line 456: perhaps your code can be simplified here if you acknowledge that a dirac is nothing more then a simple interval with both bounds closed and lower and upper being the same. Set operation work the exact same way then. For instance, EXACT_CENTER can be a number as of now or just the SimpleInterval(0.5, 0.5, Bound.CLOSED, Bound.CLOSED). Then you can proceed with it as usual :D

tomsch420 (changes requested):

PR #90, opened 2025-09-24
Connection refactor

Tigul:
Moves the content of the __post_init functions of connections to a add_to_world method which is called when adding the connection to a world. This allows to create Connections without side-effects in the world they are supposed to be added  to.

This is based on #88 and basically reverses the changes there


LucaKro (changes requested):
please let black run over this

PR #97, opened 2025-10-02
Fix dof

maltehue:
This PR fixed an issue in the DegreeOfFreedom post init functionality where, when creating a new DoF its directly registered in the symbol manager although its _world property is None until the DoF is added to the world.
This is problematic for example in the ModelSynchronizer that receives its own message, then creates a model from json with DoF that have the same name as in the current world, but then does not add them to the world because the constructed model is the same as the current world.


tomsch420 (changes requested):
Good idea, but the tests dont pass tho

ichumuh (approved):

PR #100, opened 2025-10-06
fixed merge world

ichumuh:
- merge world copies world state of the kirbied world
- connections have two transforms now: 1. for their pose, 2. for the transform that is caused by dof changes, this makes copied and merging much easier
- the old origin_expression is now a property which multiplies the two new transforms
- uncommented the copy pr2 tests, because it passes now

### DIFF
--- a/examples/regions.md
+++ b/examples/regions.md
@@ -68,7 +68,7 @@ with world.modify_world():
     leg_to_top = FixedConnection(
         parent=table_leg,
         child=table_top,
-        origin_expression=TransformationMatrix.from_xyz_rpy(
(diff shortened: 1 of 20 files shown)


Tigul (approved):
Am I seeing this correct that we can now use Connections as reference frames, also for transforming?

ichumuh:
kind of, it's not implemented, but the connections have poses now so it could be implemented. do you have a use case for that? i could add it to the pr

Tigul:
I don't really have a use-case, was just wondering if it would be possible

LucaKro (approved):

ichumuh:
so time to merge?

PR #111, opened 2025-10-24
semantic annotations now actually work

ichumuh:
turns out müslies pr didn't link the modifications to world methods


ichumuh:
alright should be done now

PR #113, opened 2025-10-24
Rip connection factories

ichumuh:
- depends on https://github.com/cram2/semantic_digital_twin/pull/111 review that first
- connections have names of dofs instead of dofs and they are mandatory arguments
- set _world of connection to init=False, if you want to make the connection create the dofs for you, you need to call with_auto_generated_dofs
- as a result removing connections often deletes all the references to a dof, so if you want to re-add it, you have to create new dofs. That's such an edge case that i don't think it matters though, so not a huge loss.
- connections are jsonable 
- deleted connection factories

what i don't like:
```                new_connection = SubclassJSONSerializer.from_json(connection.to_json())
                new_connection.parent = (
                    new_world.get_kinematic_structure_entity_by_name(
                        new_connection.parent.name
                    )
                )
                new_connection.child = new_world.get_kinematic_structure_entity_by_name(
                    new_connection.child.name
                )
                new_world.add_connection(new_connection)
```
if you from_json a connection, you the bodies are not the same ones as the ones already in the world so you have to look up by name. i could only safe the body name but that would require the from_json to get a `world` argument, which it can't. but i also don't want to keep the factories just because of that

### DIFF
--- a/src/semantic_digital_twin/exceptions.py
+++ b/src/semantic_digital_twin/exceptions.py
@@ -1,9 +1,8 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Iterable, Tuple, Union
 
-from typing_extensions import Optional, List, Type, TYPE_CHECKING, Callable
+from typing_extensions import Optional, List, Type, TYPE_CHECKING, Callable, Tuple, Union
 
 from .datastructures.prefixed_name import PrefixedName
 
@@ -131,10 +130,10 @@ class HasFreeSymbolsError(SpatialTypesError):
     Raised when an operation can't be performed on an expression with free symbols.
     """
 
-    symbols: Iterable[Symbol]
+    symbols: List[Symbol]
 
     def __post_init__(self):
-        msg = f"Operation can't be performed on expression with free symbols: {list(self.symbols)}."
+        msg = f"Operation can't be performed on expression with free symbols: {self.symbols}."
         super().__init__(msg)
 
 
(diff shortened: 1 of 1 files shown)


LucaKro (changes requested):

src/semantic_digital_twin/world_description/connections.py
```diff
@@ -339,36 +411,67 @@ def add_to_world(self, world: World):
...
-            )
-
-    def _post_init_without_world(self):
-        if any(dof is None for dof in self.passive_dofs):
-            raise ValueError(
-                "Connection6DoF cannot be created without a world "
-                "if some passive degrees of freedom are not provided."
-            )
+    @classmethod
+    def with_auto_generated_dofs(
```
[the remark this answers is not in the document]
tomsch420: also this should be an abstract method in the superclass or not?

tomsch420 (changes requested):
I like this refactoring very much, well done!

.github/workflows/build_and_deploy_doc.yml
```diff
@@ -8,6 +8,7 @@ on:
   push:
     branches:
     - main
+    - student_exercises
```
LucaKro: whops can you please delete this line? ^^

.github/workflows/exercise-test-ci.yml
```diff
@@ -0,0 +1,57 @@
...
+defaults:
+  run:
+    shell: bash -ieo pipefail {0}
+
+
+on:
+  push:
+    branches:
+      - main
+      - student_exercises
```
LucaKro: this too

LucaKro (approved):

PR #114, opened 2025-10-27
Krrood integration

tomsch420:
Removes ORMatic and EQL and integrates KRROOD instead. Also cleaned up some dirt from old changes.


src/semantic_digital_twin/adapters/ros/world_synchronizer.py
```diff
@@ -291,7 +291,9 @@ def _subscription_callback(self, msg: LoadModel):
 
         :param msg: The message containing the primary key of the model to be fetched.
         """
-        query = select(WorldMappingDAO).where(WorldMappingDAO.id == msg.primary_key)
+        query = select(WorldMappingDAO).where(
```
LucaKro: use eql here?

PR #115, opened 2025-10-28
viz markers have infinite lifetime

artnie:
In rviz the markes disappear after 100sec, so I set it to 0 to make them stay forever.

Also changed the code blocks for correct formatting when converting to .ipynb.

### DIFF
--- a/examples/loading_worlds.md
+++ b/examples/loading_worlds.md
@@ -19,7 +19,7 @@ This tutorial shows how to load a world description from a file into a `World` o
 
 First, we need to compose the path to your world file.
 
-```python
+```{code-cell} ipython3
 import logging
 import os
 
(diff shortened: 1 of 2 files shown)


src/semantic_digital_twin/adapters/viz_marker.py
```diff
@@ -97,7 +97,7 @@ def _make_marker_array(self) -> MarkerArray:
                     b=float(collision.color.B),
                     a=float(collision.color.A),
                 )
-                msg.lifetime = Duration(sec=100)
+                msg.lifetime = Duration(sec=0)
```
tomsch420: is this tested to work?

ichumuh (approved):

PR #118, opened 2025-10-31
World class cleanup

LucaKro:
### Feature: World Model Refactor, Deterministic Caching, and API Cleanups

This pull request introduces a substantial refactor of the world model, improves performance through bounded and automatically invalidated caches, and clarifies several public APIs. The primary goals are to reduce complexity in the `World` class, make interfaces harder to misuse, and improve end-to-end performance in kinematics and collision workflows.

- Theoretical Motivation: Applies SOLID principles by separating responsibilities (state, model changes, visitors), enabling safer caching via deterministic invalidation, and improving cohesion across kinematics/collision layers.
- Practical Reasoning: Faster tests and runtime, clearer method names and parameters, fewer edge-case behaviors, and easier extension for new robots/adapters.
- Related Documentation Pages: World model and state, kinematic structure and connections, spatial computations (FK/IK/Raytracer), ORM interface, ROS world sync/fetcher, semantic annotations factory, collision detection.
- Other approaches tried: Previously attempted to increase cache sizes and rely on manual invalidation; this proved fragile. Earlier refactors mixed state and behavior in `World`, raising complexity. This PR instead introduces explicit state and managers, with automatic cache invalidation keyed to model version.

---

## Highlights of This Pull-Request

- Refactored `World` to reduce complexity and move responsibilities into `WorldState` and dedicated managers (e.g., model and collision pair management).
- Introduced bounded `lru_cache` with automatic invalidation keyed to the world model version; added tests to guard cache behavior.

---

### World Model and State

- Separated model state and operations: moved parts of world-state logic from `world.py` into `world_state.py` and introduced a `WorldModelManager` for model changes. This improves cohesion and testability, and makes interfaces harder to misuse.
- Implemented automatic cache invalidation for functions using `lru_cache`. The world’s `__hash__` now incorporates a model version, ensuring caches invalidate when the model changes. A bounded cache size is enforced to avoid unbounded memory growth.
- Reduced `World` cyclomatic complexity significantly by grouping related methods and extracting visitors and managers. Tests run faster and are more stable.

References (code touched): `src/semantic_digital_twin/world.py`, `src/semantic_digital_twin/world_description/world_state.py`, `src/semantic_digital_twin/world_description/visitors.py`.

Also tried and discarded: Relying solely on larger cache sizes without invalidation, which did not robustly handle model edits.

### Kinematic Structure and Connections

- Simplified connection hierarchy; removed the passive connection type and lifted shared parameters into the superclass for clarity and maintainability.
- Clarified behavior for adding kinematic structure entities: consistent duplicate handling with explicit parameter naming.
- Improved split-chain computation and connection/body collectors via visitors.

References: `src/semantic_digital_twin/world_description/connections.py`, `src/semantic_digital_twin/world_description/visitors.py`, `src/semantic_digital_twin/world.py` (add/connection and KSE management).

Other approaches considered: Keeping passive connection as a separate subtype increased surface area without clear benefits; merging behavior reduced misuse.

### Spatial Computations (FK/IK/Raytracer)

- Moved and streamlined parts of forward kinematics into `forward_kinematics.py`; minor optimizations to reduce overhead and align with new state/caching model.
- Small cleanups in `ik_solver.py` and `raytracer.py` to align with new interfaces and caching semantics.

References: `src/semantic_digital_twin/spatial_computations/forward_kinematics.py`, `ik_solver.py`, `raytracer.py`.

### Collision Checking

- Introduced a `CollisionPairManager` to centralize collision-pair handling and reduce duplicated logic.
- Minor adjustments to `TrimeshCollisionDetector` to match refactored interfaces.

References: `trimesh_collision_detector.py`, `world.py`.

### Adapters and Integrations

- Procthor parser and ROS world fetcher/synchronizer updated to comply with the refactored API.
- URDF adapter minor adjustments.
- ORM interface (`ormatic_interface.py`) updated significantly to align with the new model/state layers and to fix a prior ORM interface issue.

References: `adapters/procthor/procthor_parser.py`, `adapters/ros/world_fetcher.py`, `adapters/ros/world_synchronizer.py`, `adapters/urdf.py`, `orm/ormatic_interface.py`.

### Semantic Annotations and Factories

- Adjusted factories to match new connection and entity behaviors; improved naming and reduced duplication.

References: `semantic_annotations/factories.py`.

### Robots

- Minor touch-ups to `AbstractRobot` and specific robot classes (`hsrb`, `pr2`, `tracy`) to integrate with refactored world and computations.

---

## Deprecated or Removed Features

- Deprecated/Removed: Passive connection type
  - Rationale: Consolidated into the base connection behavior to reduce complexity and avoid misuse. Use the unified connection interface.

- Deprecated: Implicit duplicate handling when adding kinematic structure entities
  - Replacement: Use the explicit parameter `assign_unique_name_to_duplicates` to control duplicate handling.

---

## Backwards Incompatible Changes

- Removal of passive connection subtype: Code relying on the specific subtype will need to use the unified connection base class. Suggested mitigation: Replace passive-specific references with the base `Connection` and ensure parameters are set on the superclass.
- Renamed parameters and methods for clarity:
  - `exists_ok` and `handle_duplicates` renamed to more descriptive names (e.g., `assign_unique_name_to_duplicates`). Mitigation: Search-and-replace in call sites; see updated docstrings in `world.py`.
- World hashing and caching behavior changed: Any code depending on stable `__hash__` across model edits will observe invalidation. This is by design; no action typically required.

---

## Other Changes

- Improved `__str__` for `World`.
- Exception polish and clearer custom errors for misuse cases.
- Test coverage extended for cache invalidation behavior (`test_third_party/test_lru.py`) and updated world/ROS tests.

---

Credit: Layout inspired by SciPy (https://docs.scipy.org/doc/scipy/release/1.15.0-notes.html)

### DIFF
--- a/src/semantic_digital_twin/adapters/multi_parser.py
+++ b/src/semantic_digital_twin/adapters/multi_parser.py
@@ -5,6 +5,7 @@ from dataclasses import dataclass
 import numpy
 from typing_extensions import Optional, Set, List
 
+from ..exceptions import WorldEntityNotFoundError
 from ..spatial_types.spatial_types import TransformationMatrix
 from ..world_description.geometry import (
     Box,
@@ -302,7 +303,7 @@ def parse_dof(
     """
     try:
         return world.get_degree_of_freedom_by_name(free_variable_name)
-    except KeyError:
+    except WorldEntityNotFoundError:
         if joint_builder.type == JointType.CONTINUOUS:
             dof = DegreeOfFreedom(
                 name=PrefixedName(joint_name),


src/semantic_digital_twin/world.py
```diff
@@ -238,6 +240,205 @@ def wrapper(self: World, *args, **kwargs):
...
+        by referencing bodies in the provided `world`.
+
+        :param file_path: The path to the SRDF file used for collision configuration.
+        """
+        SRDF_DISABLE_ALL_COLLISIONS: str = "disable_all_collisions"
+        SRDF_DISABLE_SELF_COLLISION: str = "disable_self_collision"
+        SRDF_MOVEIT_DISABLE_COLLISIONS: str = "disable_collisions"
+
+        if not os.path.exists(file_path):
+            raise ValueError(f"file {file_path} does not exist")
```
tomsch420: isnt there a more specific error for this?

tomsch420 (changes requested):

src/semantic_digital_twin/world_description/world_state.py
```diff
@@ -123,6 +146,20 @@ def __iter__(self) -> iter:
...
         return len(self._names)
 
+    def __eq__(self, other: Self) -> bool:
+        if self is other:
+            return True
+
+        if set(self._names) != set(other._names):
+            return False
+
+        return all(
```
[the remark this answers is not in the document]
LucaKro: if i remember correctly, the order of the data in the array may change? so I wanted to make sure that this doesnt cause weird bugs

src/semantic_digital_twin/world.py
```diff
@@ -1091,149 +1169,152 @@ def get_connection(
...
 
     def get_connection_by_name(self, name: Union[str, PrefixedName]) -> Connection:
+        connection: Connection = self._get_world_entity_by_name_from_iterable(
+            name, self.connections
+        )
+        return connection
+
+    @staticmethod
+    def _get_world_entity_by_name_from_iterable(
+        name: Union[str, PrefixedName], world_entity_iterable: Iterable[WorldEntity]
```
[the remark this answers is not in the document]
LucaKro: the reason we cant (i think) just use filter or single list comprehension is because of the typing of "name". I suggested a while ago that it should just be "prefixedname", but simon was against that, and i think you agreed with him.

I will try to improve it a little bit. if you want it to be significantly different, how would you suggest i rewrite it?

src/semantic_digital_twin/world_description/world_entity.py
```diff
@@ -475,6 +475,7 @@ def _from_json(cls, data: Dict[str, Any]) -> Self:
     "GenericKinematicStructureEntity", bound=KinematicStructureEntity
 )
 
+GenericWorldEntity = TypeVar("GenericWorldEntity", bound=WorldEntity)
```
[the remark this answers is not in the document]
LucaKro: typing for _get_world_entity_by_type_from_iterable

src/semantic_digital_twin/world.py
```diff
@@ -641,67 +653,69 @@ def add_kinematic_structure_entity(
...
-        """
-        connection.add_to_world(self)
-        for dof in connection.dofs:
-            if dof._world is None:
-                self.add_degree_of_freedom(dof)
-        self.add_kinematic_structure_entity(connection.parent, handle_duplicates)
-        self.add_kinematic_structure_entity(connection.child, handle_duplicates)
+        if lower is not None:
+            initial_position = max(lower, initial_position)
+        if upper is not None:
```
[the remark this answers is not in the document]
LucaKro: i agree

src/semantic_digital_twin/world.py
```diff
@@ -162,16 +164,16 @@ def __enter__(self):
...
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         if self.first:
-            self.world._model_modification_blocks.append(
-                self.world._current_model_modification_block
+            self.world.get_world_model_manager().model_modification_blocks.append(
+                self.world.get_world_model_manager().current_model_modification_block
             )
-            self.world._current_model_modification_block = None
+            self.world.get_world_model_manager().current_model_modification_block = None
```
ichumuh: same as above

src/semantic_digital_twin/world.py
```diff
@@ -358,56 +510,47 @@ def root(self) -> Optional[KinematicStructureEntity]:
...
-        dofs = set()
-        for connection in self.connections:
-            if isinstance(connection, ActiveConnection):
-                dofs.update(set(connection.active_dofs))
+        dofs = {
+            dof
+            for connection in self.connections
+            if isinstance(connection, ActiveConnection)
+            for dof in connection.active_dofs
+        }
```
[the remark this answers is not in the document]
ichumuh: also, it might be better to give all connections active and passive dofs, just like i did with dofs. that way to don't always need to do the isinstance

src/semantic_digital_twin/world.py
```diff
@@ -1943,63 +1933,66 @@ def compute_chain_reduced_to_controlled_joints(
...
-            ):
-                new_root = connection
-                break
-        else:
+
+        condition_for_controlled_connection = (
+            lambda conn: isinstance(conn, ActiveConnection)
+            and conn.has_hardware_interface
+            and not conn.frozen_for_collision_avoidance
+        )
```
ichumuh: why not define a proper method here? more readable than this long lambda

src/semantic_digital_twin/world.py
```diff
@@ -978,275 +1135,223 @@ def merge_world(
...
-                self.remove_kinematic_structure_entity(body)
+        rx.dfs_search(
+            self.kinematic_structure, [root_kinematic_structure_entity.index], visitor
+        )
 
-            self.semantic_annotations.clear()
-            self.degrees_of_freedom.clear()
-            self.state = WorldState()
+    # ####### Can this be removed? its not used anywhere
+    # def move_branch(
```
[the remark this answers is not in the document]
ichumuh: why? doesn't cram need that too? if you have e.g. a tray with objects on it and you pick it up, then you need this method. maybe it can be placed into some world_utils, if you dont' want it here, but i'm pretty sure we need it

src/semantic_digital_twin/world.py
```diff
@@ -1246,7 +1327,7 @@ def compute_child_kinematic_structure_entities(
             self.kinematic_structure.successors(kinematic_structure_entity.index)
         )
 
-    @lru_cache(maxsize=None)
+    @lru_cache(maxsize=512)
```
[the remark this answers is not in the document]
ichumuh: hm i see, but i could imagine that 512 is not enough for some methods, especially with the chain methods, where you have a root and a tip. for something like collision avoidance you might actually need to call it this often

src/semantic_digital_twin/world.py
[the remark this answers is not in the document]
LucaKro: actually nvm, i think this would make the code even more complicated.

im still not really happy about the union, but i get your point. ill keep it as is for now, but ill think about a solution for this

src/semantic_digital_twin/world.py
```diff
@@ -1605,52 +1636,23 @@ def compute_forward_kinematics_np(
...
-        target_frame_T_reference_frame = self.compute_forward_kinematics(
-            root=target_frame, tip=spatial_object.reference_frame
-        )
-        if isinstance(spatial_object, cas.Quaternion):
-            reference_frame_R = spatial_object.to_rotation_matrix()
-            target_frame_R = target_frame_T_reference_frame @ reference_frame_R
-            return target_frame_R.to_quaternion()
-        else:
-            return target_frame_T_reference_frame @ spatial_object
+    # May I delete this? its not used anywhere @simon
```
LucaKro: @ichumuh what about this?

src/semantic_digital_twin/world.py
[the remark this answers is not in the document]
LucaKro: sure, which number would you suggest?

src/semantic_digital_twin/world_description/connections.py
```diff
@@ -5,10 +5,12 @@
 from dataclasses import dataclass, field
 
 import numpy as np
-from typing_extensions import List, TYPE_CHECKING, Union
+from pygments.lexer import default
```
tomsch420: pretty sure you dont want this

src/semantic_digital_twin/world_description/visitors.py
```diff
@@ -0,0 +1,36 @@
...
+from typing_extensions import TYPE_CHECKING
+from .connections import ActiveConnection
+
+
+if TYPE_CHECKING:
+    from ..world import World
+    from .world_entity import Body, Connection
+
+
+class CollisionBodyCollector(rx.visit.DFSVisitor):
```
tomsch420: doc

src/semantic_digital_twin/world.py
```diff
@@ -518,60 +572,95 @@ def root(self) -> Optional[KinematicStructureEntity]:
...
+        A subset of the robot's connections that are controlled by a controller.
+        """
+        return set(
+            connection
+            for connection in self.connections
+            if connection.is_controlled
+        )
 
-    ### Add Stuff to World
+    # %% Add Stuff to World
```
tomsch420: more professional statement please

src/semantic_digital_twin/world.py
```diff
@@ -668,6 +759,7 @@ def _add_kinematic_structure_entity(
         :param kinematic_structure_entity: The kinematic_structure_entity to add.
         :return: The index of the added kinematic_structure_entity.
         """
+        self._kse_num += 1
```
tomsch420: uhhh why do we have this?

src/semantic_digital_twin/world.py
```diff
@@ -834,19 +915,21 @@ def remove_semantic_annotation(
...
-                    "The provided semantic annotation instance does not match the existing semantic annotation with the same name."
-                )
+            existing = self.get_semantic_annotation_by_name(semantic_annotation.name)
         except WorldEntityNotFoundError:
             logger.debug(
                 f"semantic annotation {semantic_annotation.name} not found in the world. No action taken."
             )
+            return
+
+        match (existing == semantic_annotation):
```
tomsch420: looks more complicated then if else imo

src/semantic_digital_twin/world.py
LucaKro: I mean the reasoning is in the docstring of the class. we need the number of bodies for the hash of the world, otherwise i need to do some hacky stuff in the viewfactories. but we dont want to have to compute  len(self.kinematic_structure.nodes()) everytime we call hash

LucaKro: I guess i can make a property "number_of_kinematic_structure_entities" or i guess "size"? then i can attach an lru cache there but also im not sure how well this would fare since its kind of referencing itself then, size properties lru cache depends on hash, but also is used in hash

src/semantic_digital_twin/world.py
```diff
@@ -518,60 +572,95 @@ def root(self) -> Optional[KinematicStructureEntity]:
...
+    def bodies_with_enabled_collision(self) -> List[Body]:
+        return [
+            b
+            for b in self.bodies
+            if b.has_collision()
+            and b.get_collision_config
+            and not b.get_collision_config().disabled
+        ]
+
+    @property
```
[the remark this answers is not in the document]
tomsch420: accepted

src/semantic_digital_twin/world.py
LucaKro: I used our tests for profiling, and did three different runs. everytime the __hash__ method of world was called 4_557_145 times:

1. call len(self.kinematic_structure.nodes()) -> 9400ms
2. use self._kse_num -> 2411ms
3. call len(self.kinematic_structure) -> 3501ms

I mistakenly used the nodes() version when i first implemented the hashing, maybe thats why it struck me as so slow. so I could maybe be convinced to use `len(self.kinematic_structure)` instead, even though its ~1 second slower.

but I also did the test you suggested to me: I spawned a world with 1000 bodies (+ 1 root), and ran len() 200 times and accessed _kse_num 200 times. this is the result:
```
len(): Time taken to calculate length 200 times: 14.78 ms. length was 1001
_kse_num: Time taken to access length using _kse_num 200 times: 5.72 ms. length was 1001
```

I ran this a few times just to vibecheck, and  len() usually takes around 15-16ms, and _kse_num 5-6ms

So I still prefer the _kse_num version. but its up to you, maybe @ichumuh and @Tigul can also contribute to this

LucaKro:
> Also make sure that the WorldMapping class is also up to date with your changes.

checked it and to me both the WorldMapping and WorldStateMapping still look fine, none of that was touched

src/semantic_digital_twin/world_description/visitors.py
```diff
@@ -0,0 +1,43 @@
...
+        if isinstance(body, Body) and body.has_collision():
+            self.bodies.append(body)
+
+    def tree_edge(self, args: Tuple[int, int, Connection]) -> None:
+        """Called for each tree edge during DFS traversal"""
+        parent_index, child_index, e = args
+        if e.is_controlled:
+            raise rx.visit.PruneSearch()
+
+class ConnectionCollector(rx.visit.DFSVisitor):
```
tomsch420: why do we even need this? Isnt this the same as just filtering the last argument of edges.

tomsch420 (changes requested):

src/semantic_digital_twin/spatial_computations/ik_solver.py
```diff
@@ -483,7 +481,7 @@ def build_goal_constraints(
         self, active_symbols: List[cas.Symbol]
     ) -> Tuple[cas.Expression, cas.Expression]:
         """Build position and rotation goal constraints."""
-        root_T_tip = self.world.compose_forward_kinematics_expression(
+        root_T_tip = self.world._forward_kinematic_manager.compose_forward_kinematics_expression(
```
ichumuh: why is the fk manager private?

ichumuh: also if the attribute is called `forward_kinematics_manager` then we could probably shorten the method names `root_T_tip = self.world._forward_kinematic_manager.compose_expression`

src/semantic_digital_twin/world.py
```diff
@@ -238,6 +235,222 @@ def wrapper(self: World, *args, **kwargs):
...
+        Notifies the system of a model change and updates necessary states, caches,
+        and forward kinematics expressions while also triggering registered callbacks
+        for model changes.
+        """
+        self.model_version += 1
+        for callback in self._model_change_callbacks:
+            callback.notify()
+
+    def get_model_modification_blocks(self):
+        return self._model_modification_blocks
```
ichumuh: why no property?

src/semantic_digital_twin/world_description/visitors.py
LucaKro: it is used in "get_connections_of_branch". i guess we could also get this by computing all descendents of a node and taking their parent connections, but this is how simon did this back then, and it utilizes rustworkx performance under the hood.

if you want I can change it, but i think its fine

src/semantic_digital_twin/world.py
```diff
@@ -294,267 +495,128 @@ class World:
...
+    def __hash__(self):
+        return hash((id(self), self._world_model_manager.model_version, self._kse_num))
 
-    def reset_temporary_collision_config(self):
-        self._temp_disabled_collision_pairs = set()
-        for body in self.bodies:
-            if body.has_collision():
-                body.reset_temporary_collision_config()
+    def __str__(self):
+        return f"{self.__class__.__name__} with {self._kse_num} kinematic structure entities."
```
ichumuh: id use f"{self.__class__.__name__} v{self.model_manager.version}.{self.state.version}."

src/semantic_digital_twin/world.py
```diff
@@ -238,6 +235,222 @@ def wrapper(self: World, *args, **kwargs):
...
+        self._disabled_collision_pairs.add(pair)
+
+
+@dataclass
+class WorldModelManager:
+    """
+    Manages the world model version and modification blocks.
+    """
+
+    model_version: int = 0
```
ichumuh: `version` would be better: `world.model_manager.version`

src/semantic_digital_twin/world_description/world_state.py
```diff
@@ -180,8 +229,78 @@ def __deepcopy__(self, memo):
...
         Create a deep copy of the WorldState.
         """
-        new_state = WorldState()
+        new_state = WorldState(_world=self._world)
         new_state.data = self.data.copy()
         new_state._names = self._names.copy()
         new_state._index = self._index.copy()
         return new_state
+
+    def add_degree_of_freedom_position(self, dof: DegreeOfFreedom):
```
ichumuh: rename to `add_degree_of_freedom`.
i don't like the `_position` subfix. there is logic to set the initial position, but there has to be. it just doesn't make sense to allow a position outside the limits.
for the type of logic used to set the limits, the doc string is good enough.

src/semantic_digital_twin/world.py
```diff
@@ -738,6 +819,85 @@ def _add_semantic_annotation(self, semantic_annotation: SemanticAnnotation):
...
+        :param kinematic_structure_entity: The kinematic_structure_entity to remove.
+        """
+        self._kse_num -= 1
+        self.kinematic_structure.remove_node(kinematic_structure_entity.index)
+        kinematic_structure_entity._world = None
+        kinematic_structure_entity.index = None
+
+    def remove_degree_of_freedom(self, dof: DegreeOfFreedom) -> None:
+        if dof._world is not self:
+            logger.debug("Trying to remove an dof that is not part of this world.")
```
ichumuh: a dof

src/semantic_digital_twin/world.py
```diff
@@ -1306,46 +1429,57 @@ def compute_chain_of_connections(
...
+        assert (
+            new_root is not None and new_tip is not None
+        ), f"no controlled connection in chain between {root} and {tip}"
+
+        # if new_root is in the downward chain, we need to "flip" it by returning its child
+        new_root_body = new_root.parent if new_root in upward_chain else new_root.child
+
+        # if new_tip is in the downward chain, we need to "flip" it by returning its parent
+        new_tip_body = new_tip.parent if new_tip in downward_chain else new_tip.child
+        return new_root_body, new_tip_body
```
ichumuh: beautiful

src/semantic_digital_twin/world.py
```diff
@@ -1364,196 +1498,127 @@ def compute_split_chain_of_connections(
...
-    def _recompute_forward_kinematics(self) -> None:
-        self._fk_computer.recompute()
+        return shortest_paths[0]
 
-    @copy_lru_cache()
-    def compose_forward_kinematics_expression(
-        self, root: KinematicStructureEntity, tip: KinematicStructureEntity
-    ) -> cas.TransformationMatrix:
+    # %% Forward Kinematics
+    def _compile_forward_kinematics_expressions(self) -> None:
```
ichumuh: shouldn't the call look like `world.forward_kinematics.compile()`?

src/semantic_digital_twin/world.py
ichumuh: why is it still deleted?

ichumuh (changes requested):

src/semantic_digital_twin/spatial_computations/ik_solver.py
LucaKro: ooh nvm i see

src/semantic_digital_twin/world.py
```diff
@@ -219,9 +213,12 @@ def wrapper(self: World, *args, **kwargs):
...
             bound_args = dict(bound.arguments)
             bound_args.pop("self", None)
-            if self._current_model_modification_block is None:
+            if (
+                self.get_world_model_manager().get_current_model_modification_block()
+                is None
+            ):
                 raise MissingWorldModificationContextError(func)
-            self._current_model_modification_block.append(
+            self.get_world_model_manager()._current_model_modification_block.append(
```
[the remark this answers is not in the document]
ichumuh: alright

src/semantic_digital_twin/world.py
```diff
@@ -294,267 +495,128 @@ class World:
...
+        self.state = WorldState(_world=self)
 
-    _temp_disabled_collision_pairs: Set[
-        Tuple[KinematicStructureEntity, KinematicStructureEntity]
-    ] = field(default_factory=set, repr=False)
-    """
-    A set of Body pairs for which collisions are temporarily disabled.
-    """
+    def __hash__(self):
+        return hash((id(self), self._world_model_manager.model_version, self._kse_num))
```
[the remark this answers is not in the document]
LucaKro: in the view factories, we create and modify new worlds. to make it easier to write new viewfactories, we wrap the _create() method in einen world context block.
but now its really annoying to get the correct root that world, since the root property previously could not be computed in a factory anymore, since its now hashed and thus would return an incorrect root.

src/semantic_digital_twin/world.py
```diff
@@ -294,267 +495,128 @@ class World:
...
-        self, entity_type: Type[GenericKinematicStructureEntity]
-    ) -> List[GenericKinematicStructureEntity]:
-        """
-        Retrieves all kinematic structure entities of a specific type from the world.
-
-        :param entity_type: The class (type) of the kinematic structure entities to search for.
-        :return: A list of `KinematicStructureEntity` objects that match the given type.
-        """
+    @property
+    def bodies_topologically_sorted(self) -> List[Body]:
```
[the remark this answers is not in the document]
LucaKro: no i dont think so, because we also want to be able to just get all kinematic_structure_entities topologically sorted. yes we could use default parameters etc and return early so we dont need to filter in those cases, but i think it just complicates the method for no reason. we also have things like "get_body_by_name" etc and its not too specific there

PR #120, opened 2025-11-04
Test for duplicated symbols after remove_connection

Tigul:
Test for bug

### DIFF
--- a/src/semantic_digital_twin/world.py
+++ b/src/semantic_digital_twin/world.py
@@ -701,9 +701,6 @@ class World:
         :param connection: The connection to add.
         """
         connection.add_to_world(self)
-        for dof in connection.dofs:
-            if dof._world is None:
-                self.add_degree_of_freedom(dof)
         self.add_kinematic_structure_entity(connection.parent, handle_duplicates)
         self.add_kinematic_structure_entity(connection.child, handle_duplicates)
 


ichumuh:
the cache of get_degree_of_freedom_by_name causes this bug.
die create_dof of omni drive makes a new dof. but in add_connection we iterate over connection.dofs. this triggers properties that use get_degree_of_freedom_by_name to get the dof from the world.
this give a cache hit and we return the old dofs which get readded to the world.
-> we now have two dofs for x and y, because the old and new connection gave them the same names

LucaKro:
actually i thought i could maybe add len(self.degrees_of_freedom) to the hash for testing. that also didnt work, but only because both connection6dof and omnidrive add the same amount degrees of freedom. when i manually add a dof before omnidrive is added to the world, and then delete it after in the same with block, then the test passes :D

LucaKro:
actually I think I found the root cause of our problem. 

Since the PR where connections only hold prefixedname references instead of the holding the degreeoffreedom references.

Before, we could create our degree of freedom seperately from our connection, and then pass it onto the connection, before then adding the whole thing to the world right? that meant that connections being added to the world may hold degrees of freedom that are not inside the world yet. hence why we had the lines:
```Python
dofs_without_world = [dof for dof in connection.dofs if dof._world is None]
for dof in dofs_without_world:
    self.add_degree_of_freedom(dof)
```

but nowadays, this test case:
```Python
def test_add_connection():
    world = World()
    b1 = Body(name=PrefixedName("b1"))
    b2 = Body(name=PrefixedName("b2"))
    connection = RevoluteConnection(parent=b1, child=b2, axis=Vector3(0, 0, 1), dof_name=PrefixedName("test"))
    with world.modify_world():
        world.add_kinematic_structure_entity(b1)
        world.add_kinematic_structure_entity(b2)
        world.add_connection(connection)
```

does not pass anymore, because it actually also assumes that the dof was already added to the world right?

definitionally `dofs_without_world = [dof for dof in connection.dofs if dof._world is None]` does not make sense anymore, because in order appear in connection.dofs, the dofs must already be in the world, since we ask the world for the dof :D

so we can just remove that part, and say if you want to manage dofs yourself then go ahead, but you also have to add it to the world yourself. 
the other option would be to add the dofs inside add_to_world().

but thats an api question, for now this bug is fixed by removing the 3 lines of code in the add_connection method

LucaKro (approved):

PR #123, opened 2025-11-06
Removed JSON serializer from World deepcopy for better perfromance

Tigul:
Removed the JSON serializer from the world deepcopy method to improve performance. For a reasonably sized world that reduces the execution time from 9.3 sec to 1.4

### DIFF
--- a/src/semantic_digital_twin/world_description/world_entity.py
+++ b/src/semantic_digital_twin/world_description/world_entity.py
@@ -4,12 +4,13 @@ import inspect
 from copy import copy, deepcopy
 
 import itertools
-from abc import ABC
+from abc import ABC, abstractmethod
 from collections import deque
 from collections.abc import Iterable, Mapping
 from dataclasses import dataclass, field
 from dataclasses import fields
 from functools import lru_cache
+from typing import ClassVar
 
 import numpy as np
 import trimesh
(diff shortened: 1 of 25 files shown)


src/semantic_digital_twin/world_description/connections.py
```diff
@@ -286,6 +287,19 @@ def jerk(self, value: float) -> None:
         self._world.state[self.dof.name].jerk = value / self.multiplier
         self._world.notify_state_change()
 
+    def __copy__(self) -> Self:
+        return self.__class__(
+            name=copy(self.name),
```
ichumuh: hm, a copy should technically do `name=self.name` no?

ichumuh (changes requested):

src/semantic_digital_twin/world_description/world_entity.py
```diff
@@ -881,6 +883,57 @@ def create_with_dofs(
...
+                parent_T_connection_expression.reference_frame.name
+            )
+        )
+
+        connection_T_child_expression = deepcopy(self.connection_T_child_expression)
+        connection_T_child_expression.child_frame = (
+            world.get_kinematic_structure_entity_by_name(
+                connection_T_child_expression.child_frame.name
+            )
+        )
```
ichumuh: nope

PR #126, opened 2025-11-09
[LMMTM] Add inertial and upgrade multiverse_parser

HoangGiang93:


HoangGiang93:
1. I only edited the files that you made, if the other tests are failed then it has nothing to do with my implementation. It failed before too ;__;
2. I will, it would be the next step, for now I want to focus on the files first
3. MuJoCo use quaternion, USD use rotation matrix, URDF use euler... Everything will be converted into either Matrix or Rotation (from scipy.spatial.tranformation import Rotation), so we don't have to think about it. You can see them in the [test case](https://github.com/HoangGiang93/semantic_world/blob/giang_dynamics/test/test_inertia/test_inertia.py#L10-L51).
4. Connection has nothing to do with inertial, it should be another file, different name, such as JointProperty. But it should be the only feature for 1DOFConnection.

HoangGiang93:
> > 1. I only edited the files that you made, if the other tests are failed then it has nothing to do with my implementation. It failed before too ;__;
> >    
> >    2. I will, it would be the next step, for now I want to focus on the files first
> >    3. MuJoCo use quaternion, USD use rotation matrix, URDF use euler... Everything will be converted into either Matrix or Rotation (from scipy.spatial.tranformation import Rotation), so we don't have to think about it. You can see them in the [test case](https://github.com/HoangGiang93/semantic_world/blob/giang_dynamics/test/test_inertia/test_inertia.py#L10-L51).
> >    4. Connection has nothing to do with inertial, it should be another file, different name, such as JointProperty. But it should be the only feature for 1DOFConnection.
> 
> 1. If you actually look at the failed test cases, they do fail because of your changes, stuff related to body not having inertial or sth. Also even If it was something unrelated to your stuff, unless the fix would require huge enough changes to warrant its own PR, it should just be fixed in this PR.
> 2. Sure thats fair, but i think it still belongs in this PR. I think the datastructures look fine, and you got toms approval as well, so it can be integrated into the parsers now i think
> 3. but that means that we will have the scipy dependency no matter what right? since we have it in testcases etc. but in that case i think having nice convenient class factory methods (from_/to_quaternion etc)  would be nice so that the user doesnt need to deal with that.
> 4. sure, but its part of the dynamics right? so I think it does have a place in this PR. feel free to add a new file then under the appropriate name you choose. I think "ConnectionProperties" is a bit too generic, can you think of a more specific name?

1. That's why I need you to provide me a template so I can fill and don't mess up your code flow. I feel like if I edit other file, then other tests can fail because of my changes. A PR with more features should only has small modification of the existing code base. I can garantee that nothing changes before and after my PR. If it failed before my PR, then it should be fixed before.
2. Ok if the datastructure looks fine then I will move forward.
3. Either we keep scipy dependency, or we remove the from_rotation and force the user to convert it into a matrix. I don't want to provide from_quaternion, from_euler... because a) it's dupplication and error-prone b) there are numerous of cases that we need to handle (quaternion can be wxyz or xyzw, euler can be implicit or explicit, xyz order or zxy oder or xyx....). However, Rotation from scipy or rotation matrix are unique.
4. Let just focus on the body properties (inertial) first, I will make another PR for joint later.

src/semantic_digital_twin/world_description/connections.py
```diff
@@ -144,6 +145,11 @@ class ActiveConnection1DOF(ActiveConnection, ABC):
     Name of a Degree of freedom to control movement along the axis.
     """
 
+    prop: JointProperty = field(default_factory=JointProperty)
+    """
+    Dynamic properties of the joint.
+    """
```
ichumuh: i'm pretty sure this should belong to degrees of freedom instead

test/test_adapters/test_multi_sim.py
```diff
@@ -236,10 +242,25 @@ def test_world_multi_sim_in_5s(self):
...
+    #     )
+    #     self.assertIsInstance(multi_sim.simulator, MultiverseMujocoConnector)
+    #     self.assertEqual(multi_sim.simulator.file_path, "/tmp/scene.xml")
+    #     self.assertIs(multi_sim.simulator.headless, headless)
+    #     self.assertEqual(multi_sim.simulator.step_size, self.step_size)
+    #     multi_sim.start_simulation()
+    #     start_time = time.time()
+    #     time.sleep(5.0)
+    #     multi_sim.stop_simulation()
+    #     self.assertAlmostEqual(time.time() - start_time, 5.0, delta=0.5)
```
ichumuh: 🤔

ichumuh:
>     3. but that means that we will have the scipy dependency no matter what right? since we have it in testcases etc. but in that case i think having nice convenient class factory methods (from_/to_quaternion etc)  would be nice so that the user doesnt need to deal with that.


What's wrong with the scipy dependency? spatial types already has it anyway

src/semantic_digital_twin/world_description/connections.py
HoangGiang93: I thought about that, but you defined a degree of freedom can be an element of a quaternion, which I can't apply the property there

src/semantic_digital_twin/world_description/connection_properties.py
```diff
@@ -0,0 +1,32 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class JointProperty:
```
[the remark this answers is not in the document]
HoangGiang93: DynamicProperty?

test/test_adapters/test_multi_sim.py
HoangGiang93: yeah I need a test with a urdf with meshes, but these meshes live in a ros package, which can't run in CI

test/test_inertial/test_inertial.py
```diff
@@ -0,0 +1,113 @@
...
+            m = PrincipalMoments.from_values(*moments_values)
+            assert_allclose(m.data, moments_values, atol=1e-12)
+
+    def test_principal_axes_properties(self):
+        for _, axes_values in moments_and_axes_values:
+            rotation_matrix = RotationMatrix.from_vectors(
+                x=Vector3.from_iterable(axes_values[:, 0]),
+                y=Vector3.from_iterable(axes_values[:, 1]),
+                z=Vector3.from_iterable(axes_values[:, 2]),
+            )
```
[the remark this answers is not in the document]
HoangGiang93: I didn't know that, thanks

HoangGiang93:
> > ```
> > 3. but that means that we will have the scipy dependency no matter what right? since we have it in testcases etc. but in that case i think having nice convenient class factory methods (from_/to_quaternion etc)  would be nice so that the user doesnt need to deal with that.
> > ```
> 
> What's wrong with the scipy dependency? spatial types already has it anyway

don't worry about it, I decided to take your RotationMatrix as the argument so the user just need to construct it.

Tigul (changes requested):

ichumuh (approved):

src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -567,10 +455,20 @@ def _convert(self, entity: ActiveConnection1DOF, **kwargs) -> Dict[str, Any]:
         joint_props = ConnectionConverter._convert(self, entity)
         assert len(entity.dofs) == 1, "ActiveConnection1DOF must have exactly one DOF."
         dof = list(entity.dofs)[0]
+        # px, py, pz, qw, qx, qy, qz = cas_pose_to_list(entity.origin) # TODO: Use actual origin
```
LucaKro: what is missing to do this?

.github/workflows/ci.yml
```diff
@@ -41,11 +41,11 @@ jobs:
         pip install -U pip && pip install -r requirements.txt && pip install -e . && pip install pytest
         pip install -e ./src/ripple_down_rules_meta
         if [[ $(python3 --version) == "Python 3.10"* ]]; then
-          PACKAGE_URL="https://nc.uni-bremen.de/public.php/dav/files/fxWQigKXPBtnXsy/?accept=zip"
-          PACKAGE_HASH="5249b21f6a6dbbb86eee9e64659cf19b9b09585ffbc888d13af9bd9c2bcce99c"
+          PACKAGE_URL="https://nc.uni-bremen.de/public.php/dav/files/5eWxp8NXt9C7BLB/?accept=zip"
```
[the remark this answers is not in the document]
HoangGiang93: I don't know any other better way ;__; blender only support 3.11, usd doesn't provide any pythonic way to extend

src/semantic_digital_twin/world_description/inertial_properties.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    Represents the inertial properties of a body. https://mujoco.readthedocs.io/en/stable/XMLreference.html#body-inertial
+    """
+
+    mass: float = 1.0
+    """
+    The mass of the body in kilograms.
+    """
+
+    center_of_mass: Point3 = field(default_factory=lambda: Point3(0.0, 0.0, 0.0))
```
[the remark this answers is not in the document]
ichumuh: what error? this should be equivalent

src/semantic_digital_twin/adapters/multi_parser.py
```diff
@@ -158,6 +169,31 @@ def parse_cylinder(cylinder: UsdGeom.Cylinder, origin_transform: TransformationM
...
+    :param plane: The UsdGeom.Mesh instance representing the plane geometry.
+    :param origin_transform: The origin transformation matrix for the plane.
+    :param color: The color of the plane.
+    :return: A Plane shape representing the parsed plane geometry.
+    """
+    size_x = plane.GetExtentAttr().Get()[1][0] - plane.GetExtentAttr().Get()[0][0]
+    size_y = plane.GetExtentAttr().Get()[1][1] - plane.GetExtentAttr().Get()[0][1]
+    size_z = plane.GetExtentAttr().Get()[1][2] - plane.GetExtentAttr().Get()[0][2]
+    if numpy.isclose(size_z, 0.0):
+        size_x = 0.0
```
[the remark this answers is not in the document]
HoangGiang93: I dont get it, what does semdt behave like?

src/semantic_digital_twin/adapters/multi_sim.py
HoangGiang93: we don't have that use case now, because in URDF all joint origins are located at the origin of the child link. There will be use cases where joint origin is not the same as the child link origin. I will keep that in mind and implement it when it's needed.
For now, yagni.

src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -917,7 +865,31 @@ def _build_shape(
         assert (
             parent_body_spec is not None
         ), f"Parent body {parent_body_name} not found."
+        if geom_props["type"] == mujoco.mjtGeom.mjGEOM_MESH:
+            mesh_entity = geom_props.pop("mesh")
```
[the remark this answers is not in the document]
HoangGiang93: ok...but it will look very ugly...

src/semantic_digital_twin/adapters/multi_sim.py
LucaKro: I dont think this is the case table.urdf for example has origin offsets for the joints

Tigul (approved):

PR #127, opened 2025-11-10
RIP symbol manager

ichumuh:
- Renames Symbol to FloatVariable, to avoid naming confusion and consistent naming with EQL. Casadi Symbols are like Variables in EQL
- Deleted Symbol Manager:
  - FloatVariables are now used by creating subclasses, which know how to resolve them. This means casadi expressions can be evaluated without a SymbolManager. just do any_casadi_expression.evaluate(), just like in EQL
  - Two FloatVariable with the same name are not unique anymore. This allows us to have two copies of the same with in the same process
  - There is still a mapping from casadi symbol to FloatVariable based on hash, which i need for .evaluate()
- FloatVariables have a PrefixedName instead of str. Makes it more consistent with everything else
- There is a bug with the caching again, where i can get old dofs in certain situations. my solution is to clear the cache of `get_degree_of_freedom_by_name` when i delete. Maybe @LucaKro has a better idea.
- _connection_T_child_expression of connections is private now, because we don't want ORM to log them. They are recreated by the post_init of connectionts


src/semantic_digital_twin/orm/model.py
```diff
@@ -99,7 +98,7 @@ class Vector3Mapping(AlternativeMapping[Vector3]):
 
     @classmethod
     def create_instance(cls, obj: Vector3):
-        x, y, z, _ = symbol_manager.evaluate_expr(obj).tolist()
+        x, y, z, _ = obj.to_np().tolist()
```
AbdelrhmanBassiouny: I would prefer to_numpy()  than to_np here tbh

src/semantic_digital_twin/spatial_types/spatial_types.py
```diff
@@ -1334,10 +1358,10 @@ def is_false_symbol(expression: Expression) -> bool:
 def logic_and(*args: ScalarData) -> ScalarData:
     assert len(args) >= 2, "and must be called with at least 2 arguments"
     # if there is any False, return False
-    if [x for x in args if is_false_symbol(x)]:
+    if [x for x in args if is_const_binary_false(x)]:
```
AbdelrhmanBassiouny: can be optimized by using any()
```python
if any(is_const_binary_false(x) for x in args)
```

src/semantic_digital_twin/spatial_computations/forward_kinematics.py
```diff
@@ -93,7 +93,7 @@ def compile(self) -> None:
                 continue
             collision_fks.append(self.child_body_to_fk_expr[body.name])
         collision_fks = cas.Expression.vstack(collision_fks)
-        params = [v.symbols.position for v in self.world.degrees_of_freedom]
+        params = [v.variables.position for v in self.world.degrees_of_freedom]
```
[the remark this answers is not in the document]
ichumuh: yes they have multiple variables, one for position, vel, acc and jerk

AbdelrhmanBassiouny (approved):

LucaKro (approved):

ichumuh:
i'm not actively working on something here, except for occasional bug fixes. the current semdt giskard that suturo is using relies on the symbol manager. if we merge it, I'd have to fix it.
I'd rather wait a little longer until my motion statechart refactoring is done. Or does anybody need this right now?

PR #131, opened 2025-11-12
Object knowledge

Tigul:
# Description 
Adds a semantic annotation for perceived objects and a few simple examples. 

Also extends the ray tracer with min and max distance parameters which allow to specify how near or far an object can be to be considered a hit.

### DIFF
--- a/src/semantic_digital_twin/world_description/world_entity.py
+++ b/src/semantic_digital_twin/world_description/world_entity.py
@@ -860,6 +860,7 @@ class Connection(WorldEntity, SubclassJSONSerializer):
         self._world = world
 
     def __post_init__(self):
+
         self.name = self.name or self._generate_default_name(
             parent=self.parent, child=self.child
         )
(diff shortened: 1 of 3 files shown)


src/semantic_digital_twin/world_description/world_entity.py
```diff
@@ -408,6 +408,18 @@ def _from_json(cls, data: Dict[str, Any], **kwargs) -> Self:
...
+    def get_semantic_annotations_by_type(
+        self, type_: Type[GenericSemanticAnnotation]
+    ) -> List[GenericSemanticAnnotation]:
+        """
+        Returns all semantic annotations of a given type which belong to this body.
+        :param type_: The type of semantic annotations to return.
+        :returns: A list of semantic annotations of the given type.
+        """
+        return list(
+            filter(lambda sem: isinstance(sem, type_), self._semantic_annotations)
```
LucaKro: please filter it like this:
```Python
[sem for sem in self._semantic_annotations if isinstance(sem, type_)]
```
1. im pretty sure its faster due to list comprehension
2. its how we have done it in all similar usecases in the codebase
3. in my opinion its more readable, because no lambda, no casting at the end, and it just reads straight forward

LucaKro (changes requested):

src/semantic_digital_twin/world_description/world_entity.py
Tigul: I don't see a benefit in doing that. 
1. Even if it would be faster the speedup is so small that it is not measurable. 
2. Fair but I think this is more readable. 
3. That is a very subjective thing. I think the way it is is more readable because the filter tells you exactly what is being done and it reads straight forward.

LucaKro (approved):

PR #141, opened 2025-11-25
Upgrades to KRROOD==2.0.1

tomsch420:


src/semantic_digital_twin/adapters/ros/messages.py
```diff
@@ -64,7 +64,7 @@ class Message(SubclassJSONSerializer, ABC):
     def to_json(self) -> Dict[str, Any]:
         return {
             **super().to_json(),
-            "meta_data": self.meta_data.to_json(),
+            "meta_data": to_json(self.meta_data.to_json()),
```
ichumuh: not `to_json(self.meta_data)`?

src/semantic_digital_twin/adapters/ros/messages.py
```diff
@@ -178,23 +174,25 @@ class WorldModelSnapshot(SubclassJSONSerializer):
...
+                "prefixed_names": to_json(self.prefixed_names),
+                "states": to_json(list(self.states)),
             },
         }
 
     @classmethod
     def _from_json(cls, data: Dict[str, Any], **kwargs) -> Self:
         state = data.get("state", {})
+
+        # Support both legacy top-level fields and the nested 'state' structure
```
ichumuh: why both?

src/semantic_digital_twin/datastructures/prefixed_name.py
```diff
@@ -25,11 +25,18 @@ def __eq__(self, other):
...
+            **super().to_json(),
+            "name": to_json(self.name),
+            "prefix": to_json(self.prefix),
+        }
 
     @classmethod
     def _from_json(cls, data: Dict[str, Any], **kwargs) -> Self:
-        return cls(name=data["name"], prefix=data["prefix"])
+        return cls(
+            name=from_json(data["name"], **kwargs),
```
LucaKro: same here, why do we need this change? why not keep it the same as before

src/semantic_digital_twin/adapters/ros/messages.py
ichumuh: which test? is it a hardcoded one? i don't see the point in supporting the old to_json implementation. shouldn't all jsons be produced be the same version of semdt?

PR #143, opened 2025-11-26
Uniqueness assumption fulfilled by UUID instead of PrefixedName now

LucaKro:
### Feature: Uniqueness assumption fulfilled by UUID instead of PrefixedName now + World-Modification Lifecycle Hardening

A broad update that streamlines name handling by removing duplicate-name workarounds, and hardens the lifecycle of world modifications.

Motivations:
- Any worldmodel has some form of uniqueness assumption, whether it be names or ids. We previously chose names, but it was hard to consistently adhere to these assumptions, especially when loading multiple objects with the same name from the database, which violated uniques of names
 
--- 

# Important Design Decision, please read
With the more enforced uniqueness assumption, deleting dofs automatically within remove_connection was causing problems. subsequently i removed it, because i also think it violated single responsibility.  One side effect of  this was, orphaned degree of freedoms would remain in the world past the with block. Luckily we already had a method, `World.delete_orphaned_dofs()`, to deal with that exact problem. I tried to make this cleanup automatic again, but this caused problems as well:
If we create a connection with the "create_with_dofs" constructor, we already add the degree of freedom, which are then for a brief while orphaned, until the connection is added. If I now clean up the dofs automatically, those dofs will be missing when adding the connection to the world, unless the connection instance was created within the same "modifies_world" block. So for now I decided to make give the user the responsibility to clean up dofs correctly, aka call the method `World.delete_orphaned_dofs()` when needed. If the world fails to validate because of this reason, then this method is also recommended as a potential fix in the Error Message.
We have a few ways to deal with this:
1. Keep it as it is now, which could take some time getting used to for some, but overall fine i think, especially since its recommended as a fix in the error message.
2. I try to re-add the automatic dof removal into remove_connection, even though i dont think it should be there. It was making problems, but also it was half way through the migration, so maybe something else was failing at this point, which caused this bug, without this bug actually being the root cause. Still, id rather not do this
3. We could also enforce say that connections, which are created with the create_with_dofs constructor must always be created within the modifies_world context block. I think this would overall be the cleanest solution, since if we say that "having orphaned dofs is not good", then having a whole range of classmethods that create orphaned dofs on purpose also isnt good. Also i think its easy to enforce that, by removing the modifies_world block from the class method, so that an context error is thrown if it is used outside of a context block.

=> Decided on option 3

## Highlights of This Pull-Request

- Strict name/identity handling: removal of duplicate-name helpers and reliance on explicit `PrefixedName` usage.
- Parser and adapter refinements (Procthor, ROS, URDF) to keep pace with naming/ID changes and ensure consistent synchronization.
- there are now two World.get_..._by_name versions. one in singular, which behaves as before (throws error if multiple are found), and one in plural, which returns a list of matched names, as this is now possible.
- If we want to  identify one unique entity, we can now also search by ID. but this is mostly used internally or for deserialization, since if we have the ID, we most likely already have the object

---

## Deprecated or Removed Features

- Deprecated/Removed: Duplicate-name utilities `assign_unique_name_to_duplicates` and `skip_duplicates` removed. Reason: Implicit renaming created ambiguity in identity and broke predictable reasoning. Replacement: Use explicit `PrefixedName` and enforce unique naming at creation time.

---

## Backwards Incompatible Changes

- Naming behavior: Removal of implicit duplicate-name helpers. Impact: Worlds that previously depended on silent renaming will now need explicit, deterministic naming. Mitigation: Provide `PrefixedName` upfront and ensure uniqueness at source.
- UUID adjustments: The identity model has been tuned to ensure stability and clarity (commit references indicate UUID-related changes). Impact: Stored references or assumptions about previous UUID formats can break. Mitigation: Regenerate or migrate persisted data; treat UUIDs as opaque identifiers.

---

## Other Changes

- Test robustness: Fixed deepcopy tests to ensure true deep-copy semantics and caught cases where blocking operations modified state unexpectedly. Tests adjusted across `test_*` to align with new requirements.
- Minor quality improvements across adapters and reasoning components to reduce hidden state and increase code clarity.

---

Credit: Layout inspired by SciPy (https://docs.scipy.org/doc/scipy/release/1.15.0-notes.html)


src/semantic_digital_twin/world.py
```diff
@@ -1085,51 +1032,169 @@ def get_degree_of_freedom_by_name(
...
             case str():
-                pass
+                return [
+                    world_entity
+                    for world_entity in world_entity_iterable
+                    if world_entity.name.name == name
+                ]
+
+    def get_degree_of_freedom_by_id(self, id: UUID) -> DegreeOfFreedom:
+        return self._get_world_entity_by_hash_from_iterable(hash(id), self.degrees_of_freedom)
```
AbdelrhmanBassiouny: Why no just by the id? Why hash?

src/semantic_digital_twin/world_description/world_state.py
```diff
@@ -105,54 +107,57 @@ def _notify_state_change(self) -> None:
...
-        self, name: PrefixedName, value: np.ndarray | WorldStateView
+        self,dof_or_uuid: Union[DegreeOfFreedom, UUID], value: np.ndarray | WorldStateView
     ) -> None:
         if isinstance(value, WorldStateView):
             value = value.data
         arr = np.asarray(value, dtype=float)
         if arr.shape != (4,):
             raise ValueError(
-                f"Value for '{name}' must be length-4 array (pos, vel, acc, jerk)."
+                f"Value for '{dof_or_uuid}' must be length-4 array (pos, vel, acc, jerk)."
```
AbdelrhmanBassiouny: Custom exception

src/semantic_digital_twin/world_description/world_entity.py
```diff
@@ -53,8 +56,8 @@
 id_generator = IDGenerator()
 
 
-@dataclass(unsafe_hash=True, eq=False)
-class WorldEntity(Symbol):
+@dataclass(eq=False)
+class WorldEntity(Symbol, ABC):
```
ichumuh: why is world entity ABC?

AbdelrhmanBassiouny (changes requested):
Nice work 👌

ichumuh (commented):
i agree that create_with_dofs should require a with block. i actually thought it was already this way before

src/semantic_digital_twin/world_description/world_entity.py
LucaKro: why not? we never want to instantiate a "WorldEntity" directly? its just the abstract base class (:p) for classes whose instances may belong to a world

ichumuh: there are no abstractmethods, so this doesn't change anything. you can still instantiate it.

src/semantic_digital_twin/adapters/procthor/procthor_parser.py
```diff
@@ -642,9 +633,7 @@ def import_objects(self, world: World, objects: List[Dict]):
                 child=obj_world.root,
                 parent_T_connection_expression=procthor_object.world_T_obj,
             )
-            world.merge_world(
-                obj_world, obj_connection, assign_unique_name_to_duplicates=True
-            )
+            world.merge_world(obj_world, obj_connection)
```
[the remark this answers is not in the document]
LucaKro: yeees

tomsch420 (changes requested):
Generally good, just go for the automatic cleanup of dofs as discussed during lunch.

src/semantic_digital_twin/world.py
```diff
@@ -794,44 +740,26 @@ def _add_degree_of_freedom(self, dof: DegreeOfFreedom) -> None:
...
-        :param actuator: The actuator to be added to the system.
-        :return: None
-        """
-        actuator._world = self
-        self.actuators.append(actuator)
-
-    def add_semantic_annotation(
-        self, semantic_annotation: SemanticAnnotation, skip_duplicates: bool = False
-    ) -> None:
+    def add_semantic_annotation(self, semantic_annotation: SemanticAnnotation) -> None:
```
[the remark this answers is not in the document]
LucaKro: outdated doc, fixed

src/semantic_digital_twin/world.py
LucaKro: because this way we can use this method to find all world entities by hash, if we at some point need to find an exact connection or semantic annotation we can use that, and it wasnt more complicated or more work that using id, so i dont think its yagni

src/semantic_digital_twin/world_description/world_state.py
```diff
@@ -105,54 +107,57 @@ def _notify_state_change(self) -> None:
...
             self.data = np.hstack((self.data, new_col))
 
-    def __getitem__(self, name: PrefixedName) -> WorldStateView:
-        if name not in self._index:
-            self._add_dof(name)
-        idx = self._index[name]
+    def __getitem__(self, dof_or_uuid: Union[DegreeOfFreedom, UUID]) -> WorldStateView:
+        dof_id = dof_or_uuid.id if isinstance(dof_or_uuid, DegreeOfFreedom) else dof_or_uuid
+        if dof_id not in self._index:
+            self._add_dof(dof_id)
```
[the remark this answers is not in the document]
LucaKro: what about setitem? only allow to modify existing keys? i think that makes sense tbh

Tigul (approved):

ichumuh (approved):

src/semantic_digital_twin/adapters/world_entity_kwargs_tracker.py
```diff
@@ -88,19 +89,19 @@ def add_kinematic_structure_entity(
...
             return True
         except KinematicStructureEntityNotInKwargs:
             return False
 
     def get_kinematic_structure_entity(
-        self, name: PrefixedName
+        self, id: UUID
     ) -> KinematicStructureEntity:
         """
         Retrieve a kinematic structure entity by its name.
```
AbdelrhmanBassiouny: fix doc, `by its name` -> `by its id`

src/semantic_digital_twin/orm/utils.py
```diff
@@ -22,7 +22,7 @@ def persistent_database_available() -> bool:
         return False
 
     try:
-        engine = create_engine(f"mysql+pymysql://{semantic_digital_twin_database_uri}")
+        engine = create_engine(f"{semantic_digital_twin_database_uri}")
```
AbdelrhmanBassiouny: no need to wrap it in an f"" string

src/semantic_digital_twin/world_description/world_entity.py
```diff
@@ -1096,3 +1133,51 @@ def _attr_values(
...
+
+    def add_dof(self, dof: DegreeOfFreedom) -> None:
+        """
+        Adds a degree of freedom to this actuator.
+
+        :param dof: The degree of freedom to add.
+        """
+        self._dofs.append(dof)
+
+    def __hash__(self):
```
AbdelrhmanBassiouny: why not put this hash in WorldEntityWithID?

src/semantic_digital_twin/world_description/world_modification.py
```diff
@@ -340,6 +332,29 @@ def _from_json(cls, data: Dict[str, Any], **kwargs) -> Self:
         return cls(actuator=Actuator.from_json(data["actuator"], **kwargs))
 
 
+@dataclass
+class RemoveActuatorModification(WorldModelModification):
```
AbdelrhmanBassiouny, line 336: Modification in the name feels duplicate with the parent WorldModelModification

src/semantic_digital_twin/world.py
```diff
@@ -672,113 +675,57 @@ def _add_connection(self, connection: Connection):
 
         :param connection: The connection to be added to the kinematic structure.
         """
-        connection._world = self
+        connection.add_to_world(self)
```
AbdelrhmanBassiouny, line 678: Why does connection add itself to the world? Only the world can add connections to itself.

AbdelrhmanBassiouny: otherwise you would have a lot of clutter code because of delegating world methods to world entities

src/semantic_digital_twin/world.py
```diff
@@ -794,103 +741,119 @@ def _add_degree_of_freedom(self, dof: DegreeOfFreedom) -> None:
...
+        parent_index = connection.parent.index
+        child_index = connection.child.index
+        if parent_index is not None and child_index is not None:
+            try:
+                self.kinematic_structure.remove_edge(
+                    parent_index, child_index
+                )
+            except NoEdgeBetweenNodes:
+                pass
+        connection.remove_from_world()
```
AbdelrhmanBassiouny, line 856: same with the add_to_world, but maybe a change in another pr

.github/workflows/ci.yml
```diff
@@ -40,20 +40,6 @@ jobs:
         git submodule update
         pip install -U pip && pip install -r requirements.txt && pip install -e . && pip install pytest
         pip install -e ./src/ripple_down_rules_meta
-        if [[ $(python3 --version) == "Python 3.10"* ]]; then
-          PACKAGE_URL="https://nc.uni-bremen.de/public.php/dav/files/9XcXtFD3fDmpMTB/?accept=zip"
```
tomsch420: so no multiverse anymore?

src/semantic_digital_twin/adapters/multi_parser.py
tomsch420, line 1: if this file is empty now delete it

src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -244,6 +272,24 @@ def _from_json(cls, data: Dict[str, Any], **kwargs) -> Self:
             f" Use TriangleMesh instead."
         )
 
+    @classmethod
+    def from_spec(
```
tomsch420: i find the method name confusing

tomsch420: and also no abbrevations

src/semantic_digital_twin/world_description/world_modification.py
tomsch420: i think its fine

tomsch420 (changes requested):

src/semantic_digital_twin/world_description/world_entity.py
LucaKro: actually, it is :D i deleted this one now

src/semantic_digital_twin/world_description/world_modification.py
LucaKro: I think its correct like this. this is different from methods that have the name of its class inside.

Here its also about user api. and RemoveActuator is not descriptive enough in my opinion. RemoveActuatorModification is actually what it is, and its in line with our naming of Modifications

src/semantic_digital_twin/world.py
LucaKro: because connections have to do extra stuff when being added to the world. This is just me continuing that pattern for consistency.If you think that this should be done differently we could look into it, but i think this is out of scope for this PR

src/semantic_digital_twin/world.py
LucaKro: yes, and we need to discuss it first. I think remove_from_world could easily be removed, but i did it for consistency with add_to_world. add_to_world may be more work, but we need to revisit the reason why we did it i think

.github/workflows/ci.yml
LucaKro: this was not from me, this was from another PR (i think giangs), but github is displaying changes weirdly lately.

I think multiverse is now a normal pip requirement since the USD dependency is gone or sth

src/semantic_digital_twin/adapters/multi_parser.py
LucaKro: did

src/semantic_digital_twin/world_description/geometry.py
LucaKro: also not my PR, was giangs. I can change it tho

LucaKro: changed it to `from_file`, same name we used for the same method in the URDF parser

src/semantic_digital_twin/world.py
AbdelrhmanBassiouny: ok fair

AbdelrhmanBassiouny (approved):

Tigul (approved):

test/test_adapters/test_multi_sim.py
HoangGiang93, line 1: it works on my PC

LucaKro: its flaky, most of the time it seems to work, but sometimes it doesnt

PR #144, opened 2025-11-27
SymbolicType and thus float variables and expressions are not symbols

ichumuh:
you can now query for expressions and float variables


AbdelrhmanBassiouny (approved):

ichumuh:
this needs a new krrood version to pass

LucaKro (approved):
approve when the tests pass

PR #145, opened 2025-11-27
Add Perceivable Semantic Annotations for RoboKudo

fmuehlis:
Added some Semantic Annotations that are used for the Semantic Digital Twin demo in RoboKudo. Are these generalized enough?

For example:
- "Milk" refers to a carton containing milk
- "Salt" refers to a packet of salt

I'm not quite sure especially about these two, but since raw salt and milk (the liquid) would rather be "stuff" and not a "thing" I think it should be fine?

The various objects are to be seen in this image:
image


LucaKro (changes requested):

src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul, line 1: Double check with the semantic annotations in adapters.procthor.procthor_semantic_annotations to avoid duplications

fmuehlis:
I have removed the duplicated Cup definition. I also found a duplicated Milk:
https://github.com/cram2/semantic_digital_twin/blob/main/src/semantic_digital_twin/adapters/procthor/procthor_semantic_annotations.py#L463

Although it inherits from `Cuttlery` is that correct?

Edit: Tests seem to fail due to circular imports. So it is not possible to let `Candy` inherit from `Food`.

LucaKro:
> I have removed the duplicated Cup definition. I also found a duplicated Milk: https://github.com/cram2/semantic_digital_twin/blob/main/src/semantic_digital_twin/adapters/procthor/procthor_semantic_annotations.py#L463
> 
> Although it inherits from `Cuttlery` is that correct?
> 
> Edit: Tests seem to fail due to circular imports. So it is not possible to let `Candy` inherit from `Food`.

circular imports usually tell you that the structure of your program needs work. try moving the class causing the circular import

fmuehlis:
Well this is not my program. Hence I don't really know how you would want this to be resolved and would be thankful for some guidance. For example
I could move all annotations from procthor adapter to where all other semantic annotations are, but I don't know why they are in the adapter in the first place. I can also create a new adapter module for RoboKudo and put all related semantic annotations there if that is the structure you want to go for?

tomsch420:
These semantic annotations are very misleadiing. If its a container for milk and not the substance itself, call it milkcontainer. Also the docstrings dont need to all start with "a semantic annotation for". Keep it concise. Semantic annotations are about structuring the world as detailed as possible and createa precise vocabulary for everything

Tigul:
I think the distinction between a milk and a milk container is unnecessary and just introduces overhead in using these. When I want to talk about a milk it is already implied that the Milk container is meant. 

@fmuehlis Moving the semantic annotations from adapters.procthor to semantic_annotations should be the best solution, there is no reason why they should be in the procthor adapter in the first place.

src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -158,3 +160,465 @@ def doors(self) -> Iterable[Door]:
...
+class Milk(Container, IsPerceivable):
+    """
+    A container of milk.
+    """
+    ...
+
+
+# TODO: Which Milk to keep?
+# @dataclass(eq=False)
+# class Milk(Cuttlery, IsPerceivable): ...
```
fmuehlis, line 184: I am not sure which Milk to keep here. Does it really make sense to call Milk "Cuttlery"?

fmuehlis:
>These semantic annotations are very misleadiing. If its a container for milk and not the substance itself, call it milkcontainer. Also the docstrings dont need to all start with "a semantic annotation for". Keep it concise. Semantic annotations are about structuring the world as detailed as possible and createa precise vocabulary for everything

> I think the distinction between a milk and a milk container is unnecessary and just introduces overhead in using these. When I want to talk about a milk it is already implied that the Milk container is meant.

@Tigul @tomsch420 What could be a solution here? Is there a need to have a distinction between the substance Milk and a container for it?

In case there is not, would it still be correct then to inherit from Container then? Or would I rather also use `HasBody`, `Food` or even introduce some other superclass `Drinkable` or `Liquid`? Is the same implication also true for the `Noodles`, `Cereal` and `Salt` containers?

> @fmuehlis Moving the semantic annotations from adapters.procthor to semantic_annotations should be the best solution, there is no reason why they should be in the procthor adapter in the first place.

I have moved the procthor annotations to the same file as the other annotations. I also adjusted the docstrings to be more concise.

PR #150, opened 2025-11-30
Faster state synchronization

ichumuh:
- told junie to make compute_state_changes faster, it told me it was buggy, fixed it, made it much faster and wrote tests.
- fixed a bug where paused statecallbacks would be started after receiving a world update
- I also refactored CompiledFunction a little bit, which allows me to do evil stuff in giskard for more speed.


src/semantic_digital_twin/spatial_types/spatial_types.py
```diff
@@ -240,9 +240,23 @@ def _setup_constant_result(self) -> None:
         self._function_evaluator()
         self._is_constant = True
 
+    def bind_args_to_memory_view(self, arg_idx: int, numpy_array: np.ndarray) -> None:
+        """
+        Binds the arg at index arg_idx to the memoryview of a numpy_array.
+        If your args keep the same memory across calls, you only need to bind them once.
```
LucaKro, line 246: are we sure this is the case? if not couldnt this introduce super hard to debug errors?

ichumuh: the tests verify it, but it would probably be a good idea to add tests that verify that the world state keeps reusing the same memory.
and yes it can create bugs that are hard to debug, but the world state already kind of is, due to the optimizations. Basically the world state is already a single numpy array that is being reused to avoid memory allocations on each update. I needed that to save time. 
Only difference is that I realized that since the world state reuses the same memory, i don't need to set the args every time. This actually saves like 10-30% per function call in giskard, depending on how many args the function has.

LucaKro: I think its appropriate to add a test to verify the world state reuses memory just to make sure. other than that its good :+1:

LucaKro (approved):

PR #151, opened 2025-12-01
collision avoidance fix

ichumuh:


LucaKro (approved):
removing the lru caches here is a quickfix, we should fix the underlying problem instead. but for now its fine because of the monorepo merge

ichumuh:
nvm we can keep it

